PR #16, opened 2017-02-07
Update CoExtendedAttributesController.php

mheder:
The TIMESTAMP column does not work without DEFAULT in some recent versions of mysql. This causes a nasty, unexplained DB error in COmanage when attempting to create extended attributes.

### DIFF
--- a/app/Controller/CoExtendedAttributesController.php
+++ b/app/Controller/CoExtendedAttributesController.php
@@ -242,8 +242,8 @@ class CoExtendedAttributesController extends StandardController {
         $sql = "CREATE TABLE " . $cotable . " (
                 id SERIAL PRIMARY KEY,
                 co_person_role_id INTEGER REFERENCES " . $this->CoExtendedAttribute->tablePrefix . "co_person_roles(id),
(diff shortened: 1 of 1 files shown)


boshrin:
Which version of MySQL does this patch apply to?

Unfortunately I don't think we can merge as is since [prior to v5.6.5](https://dev.mysql.com/doc/relnotes/mysql/5.6/en/news-5-6-5.html) MySQL does not allow multiple default current timestamps in the same table.

    Server version: 5.5.52-0+deb8u1 (Debian)
    mysql> create table test (id serial primary key,
                              created timestamp default current_timestamp,
                              modified timestamp default current_timestamp);
    ERROR 1293 (HY000): Incorrect table definition; there can be only one TIMESTAMP column with CURRENT_TIMESTAMP in DEFAULT or ON UPDATE clause`

muisit:
I am having difficulties reproducing the behaviour I originally detected. It probably was a MariaDB error, but I think I ran into it during provisioning on a platform I cannot recall at the moment. The code as it is seems to work fine on the platform I am currently using, so I am closing this PR.

PR #17, opened 2017-03-17
Fix DatabaseShell for PHP 7 and MySQL (CO-1433)

skoranda:


boshrin:
Not super thrilled with the hardcodedness of this, but there's probably not a better way to handle this at the moment...

PR #19, opened 2017-05-11
CO-1427 Fix table name

tomzeller:
Fix typo in extended attributes table name. Fixes PostgreSQL, but probably not MySQL.


boshrin:
Ah, good catch. A couple of comments...

1. This should probably be modified to use $cotable, which is defined towards the top of checkWriteDependencies.
2. CO-1427 is targeted for v2.0.1, so the pull request should be made against hotfix-2.0.x, not develop.
3. Most of the code in checkWriteDependencies should be refactored into the Model (this code is really, really old), but that's out of scope here. Mostly noting for future reference.

PR #21, opened 2017-05-27
Configuration submenu getting long (CO-1455)

spartanicus:
Moved new page to co_dashboards and removed co_configuration controller along with its view


spartanicus:
The lang key has been renamed and moved to its own section for dashboards "db.configuration" - this seemed more appropriate than another reasonable choice of "op.dashboard.configuration".  A case could be made that at some point "op.dashboard.select" should be moved down to join it - as "db.co.firstpage" or similar.

spartanicus:
The key has been changed - and now sits with the other "op.dashboard" key.

PR #22, opened 2017-05-31
CO-1362 Submit form once

tomzeller:
Selecting an Organizational Identity Source automatically submits the form so that clicking submit would therefore submit twice and likely break CakePHP's csrf checks.


spartanicus:
This is a good catch and a similar change should be made to all forms that use this behavior both for the reason stated here and because it creates accessibility issues.

PR #23, opened 2017-06-08
CO-1441 Add note to form field

tomzeller:
Add reminder to not use a reserved word.


boshrin:
While the key fd.lan.desc is not used anywhere else, it seems like it could be ("lower alpha numeric"). Perhaps we should add a new key specifically for this scenario? eg fd.lan.res.desc?

tomzeller:
Add new localization key as suggested.

PR #24, opened 2017-06-24
Fix handling of empty attribute from LDAP schema plugin (CO-1492)

skoranda:


skoranda:
I have pushed another commit that moves the modify check into assembleAttributes() where the attributes from the schema plugin are first returned from the plugin.

PR #28, opened 2017-11-23
Fix on ServiceToken submenu

HarryKodden:
The Service Token SubMenu does not show.  Now it shows inside the usermenu.
Additionally: The link to the Service Tokens was incorrect, it was missing the 'copersion:' element


boshrin:
Actually hotfix-3.0.x is no longer open. The PR branch should be set to hotfix-3.1.x if such a branch exists when ready (we won't create it for a few days), develop otherwise.

HarryKodden:
@boshrin Please elaborate: hotfix-3.0.x was no longer open (as you mentioned) but hotfix-3.1.x still does not exist. Please provide met with instructions on where to promote the fix.

boshrin:
@HarryKodden Since we have not yet released 3.1.0, the appropriate branch to submit the PR against is "develop". This should be true for a few more days.

boshrin:
Replaced by pull request #40 .

PR #31, opened 2017-12-18
CO-1544 Allow Terms and Conditions to be ordered

tomzeller:
https://bugs.internet2.edu/jira/projects/CO/issues/CO-1544

### DIFF
--- a/app/Config/Schema/schema.xml
+++ b/app/Config/Schema/schema.xml
@@ -1822,6 +1822,7 @@
       <constraint>REFERENCES cm_cous(id)</constraint>
     </field>
     <field name="status" type="C" size="2" />
+    <field name="ordr" type="I" />
     <field name="created" type="T" />
     <field name="modified" type="T" />
     <field name="co_terms_and_conditions_id" type="I">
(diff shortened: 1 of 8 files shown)


boshrin:
Also this should be submitted against develop, not hotfix-*, as it's really an RFE not a bug.

tomzeller:
Looks like (1) is because I copied [6a2655e178](https://github.com/Internet2/comanage-registry/commit/6a2655e178eda70dba49119d6a5d0eca7e282860#diff-a92f465c1299be94e1443ff52de3cbedR89).

Should I submit a PR for that class too ?

tomzeller:
And app/View/CoTermsAndConditions/index.ctp:77 :

`<th><?php print $this->Paginator->sort('order', _txt('fd.tc.order')); ?></th>`

should be :

`<th><?php print $this->Paginator->sort('ordr', _txt('fd.tc.order')); ?></th>`

tomzeller:
Benn, over to you for review.

boshrin:
Re 6a2655e178, there's a lot of old code that shouldn't be copy/pasted :)

PR #41, opened 2018-02-22
Changed link in the finalization notification to the person informati…

muisit:
This fix changes the source URL in the finalization notification message from a view of the petition to the canvas view of the related CoPerson.
Chances are high that the newly enrolled user has no privileges to view his/her own petition, so sending this link will cause a permission denied error. Instead, directing the new user to their registered data will allow them to immediately adjust any wrong or missing information and will never cause a permission denied error.


boshrin:
I see the source of the confusion, but I'm not yet convinced this is the correct approach. The notification registration URL should be closely correlated to the source, so changing it to the CO Person canvas may not make sense. It might make more sense to allow enrollees to see their petitions in the first place, or to offer some sort of customized message otherwise.

Can you open a ticket in JIRA so we can track this there?

boshrin:
BTW, as another question, can you use a Message Template for approval notification? That's typically how we'd recommend customizing these notifications.

https://spaces.internet2.edu/display/COmanage/Message+Templates

muisit:
I think you can work around this issue in the current setup by specifying a notification message. But the issue is that by default the application sends out an invalid link. By default it should 'do the right thing'. Perhaps it should never send out a notification to the petition user to start with. You can separate the 'notify on finalization' in two options: 'notify admin on finalization' and 'message user on finalization' with a simple default template for the latter option.

boshrin:
As per f2f conversation, closing this PR as minimally we can't change the registration URL for notifications this way. Further discussion in CO-1602.

PR #42, opened 2018-02-23
CO-1599 Refresh model tests

tomzeller:
Existing tests are now working except for CouTest->testBeforeDelete().


tomzeller:
Updated @since as requested.

PR #43, opened 2018-02-26
Fix for situations where changes in COs, groups and identities (e.g. …

muisit:
This is a fix I pushed to the SURFnet fork of COmanage to work around an issue where session stored variables were changed due to petition finalization, but the session was not refreshed. This caused authenticated users to miss (ie: not have) a user menu with a logout option (which refreshes their session variables), effectively locking them into a page, even though they were properly accepted into a CO. 
This is predominantly a problem for first-time enrollment. If you enroll for your second CO, you get a user menu based on your first CO anyway, so you can logout.

This fix moves the session-storage code from UserController to a LoginComponent and invokes the process method both from the UserController after login as well as on each (regular, non-REST) page load through StdController.
This, of course, causes additional overhead on each page load and I can only assume that storing these values in the Session object was meant to prevent this overhead. 

I could not find a related Jira issue for this problem, though one might exist.

Alternatives to this fix could be:
- having a 'flag' option for a user record to indicate the session variables need to be re-established, which could be set upon changing a users groups and COs
- rereading the session variables only once every X page loads or Y seconds instead of on every page load
- displaying an explicit session-refresh option on each page (re-login), empowering users to overcome this bug

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -47,7 +47,8 @@ class AppController extends Controller {
                              'Role',
                              'Security',
                              'Session',
-                             'Paginator');
+                             'Paginator',
+                             'Login');
   
   // We should probably add helpers here instead of in each Controller. To do so,
   // make sure to define the default Html and Form helpers (and Flash).
@@ -216,6 +217,9 @@ class AppController extends Controller {
       } else {
         $this->set('vv_tz', date_default_timezone_get());
       }
+
+      // (Re)set the session groups, COs and identities related to this user.
+      $this->Login->process();
 
       // Before we do anything else, check to see if a CO was provided.
       // (It might impact our authz decisions.) Note that some models (eg: MVPAs)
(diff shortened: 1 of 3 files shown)


boshrin:
I'm a bit confused by the premise here, since a user logout/login cycle has been required after an enrollment for quite some time. What is the sequence of actions that precedes this issue?

FWIW, the related JIRA is probably [https://bugs.internet2.edu/jira/browse/CO-1392](CO-1392), and a lot of this code is going to need to be refactored for the framework migration (v4.0.0).

muisit:
I thought I had replied to this, but I can't find it anywhere, so it is probably just in my head.
Anyway, stepping through the CoPetitionController code, I think this forced logout is only effective for email confirmation and/or self-registration flows. The two places where a redirect to auth/logout is made are wrapped in conditionals, so I cannot imagine that a login/logout cycle is mandatory during enrollment.

Anyway, I think the sequence is an invitation flow:
- user receives invitation, clicks on review link, logs in and and accepts the petition. Session is started
- administrator accepts new petition, finalization notification is sent to new user
- new user receives email, clicks on either of the two links and ends up in the COmanage application, but still in the old session without the new OrgIdentity, CO or COGroup information. 

There is no login/logout cycle in between accepting the petition and the user revisiting the application.

PR #45, opened 2018-03-09
CO-1605: implementation of a text body field for Terms&Conditions to …

muisit:
This PR implements a fix for CO-1605: adding a body field to the T&C to override the URL setting. It implements a local URL that displays the T&C content in raw form. 
There is only a sanity filter on the T&C body and no filter on its output. I assume there are possibilities for administrators with bad intentions to display vulnerable javascript exploits in this way, but I am not sure if that is an issue. Alternatively, the raw view could strip tags to only allow some well-known markup tags (p, hN, em, i, b, div, span,body,html,img(?),a, etc.)

Additionally, support could be added to display non-HTML content by wrapping it in html tags and applying nl2br()

More additionally, HTML content could be stripped and only text content supported, but that would reduce the layout possibilities. That could be countered by supporting a form of markdown, but that might go too far at this point.

This PR also fixes a syntax issue in the LdapProvisioner around the original empty() check on the URL. A bracket was missing after the empty().

I tested LdapProvisioning of a voPerson objectclass. I tested checking the T&C at login. I tried testing the T&C during enrollment, but encountered the issue that for invitation flows, the administrator initiating the flow needs to agree to the T&C. I think accepting the T&C during enrollment should be moved to the phase where you accept your enrollment attributes. That would circumvent this specific issue and it could be argued that you don't need to accept the T&C to start the enrollment, but only to finish (or 'definitely accept') it.

### DIFF
--- a/app/Config/Schema/schema.xml
+++ b/app/Config/Schema/schema.xml
@@ -2064,6 +2064,7 @@
     </field>
     <field name="description" type="C" size="256" />
     <field name="url" type="C" size="256" />
+    <field name="tc_body" type="X2" />
     <field name="cou_id" type="I">
       <constraint>REFERENCES cm_cous(id)</constraint>
     </field>
--- a/app/Controller/CoTermsAndConditionsController.php
+++ b/app/Controller/CoTermsAndConditionsController.php
@@ -195,10 +195,35 @@ class CoTermsAndConditionsController extends StandardController {
     // Modify ordering for display via AJAX
     $p['reorder'] = ($roles['cmadmin'] || $roles['coadmin']);
 
+    // A raw view of the T&C content is always allowed
+    $p['raw_view']=TRUE;
+
     $this->set('permissions', $p);
     return $p[$this->action];
   }
-  
+
+  /**
+   * Raw view of the T&C
+   * Necessary for CO-1605, to enable viewing of T&C through the existing
+   * framework, as well as storing the relevant T&C as URL in the LDAP
+   * provisioner.
+   *
+   * @since  COmanage Registry v3.2.0
+   * @param  Integer $id CO Terms and Agreements identifier
+   * @return void
+   */
+
+  public function raw_view($id) {
+    $args = array();
+    $args['conditions'][$this->modelClass.'.id'] = $id;
+
+    $tc = $this->CoTermsAndConditions->find('first', $args);
+
+    print ($tc[$this->modelClass]['tc_body']);
+
+    exit(0);
+  }
+
   /**
    * Pull T&C for review
    * - precondition: The named parameter 'copersonid' is populated with the target CO Person ID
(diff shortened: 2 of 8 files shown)


boshrin:
In order to merge this, we'll need several changes, most of which are to better fit the existing coding style:

1. `tc_body` should just be `body`. I realize other tables use the form foo_body, but going forward we're trying to reduce redundant naming, and the "tc" is already implied by the table name.
2. For consistency, the schema field configuration should be `type="X" size="4000"`. I don't think X2 actually gets us anything, as we've tested with UTF8 and haven't needed to flag anything specially.
3. Instead of `raw_view` let's name it `display`, which is consistent with dashboard widgets, and somewhat more accurate (since we'll render the HTML, not display the source).
   1. In `raw_view`, is there a reason to explicitly `exit(0)`? We typically let the process "naturally" exit except in rare circumstances.
   2. In the function comment, we don't need to reference CO-1605. JIRA references in the code are typically only for work still to do, or particularly complicated discussions that might be useful to preserve.
4. Per the [coding guidelines](https://spaces.at.internet2.edu/display/COmanage/COmanage+Coding+Style), `$p['raw_view']=TRUE` should be `$p['raw_view'] = true` and `'required'=>FALSE` should be `'required' => false`
5. Rather than have logic repeated in the CoPetition, LdapProvisioner and CoTermsAndConditions code, let's add a `beforeSave` callback to set the `url` value to the appropriate display URL if `body` is not empty. This will also ensure consistency.
6. Can we generalize the validation rule and move it to AppModel (where our other custom validation rules are)? eg: `validateOneOfMany` or something that accepts a list of fields and requires that exactly one of them has a value.
   1. Also, the debug calls should be removed.
   2. We should try to avoid hardcoding unlocalized text strings if we can, though some other Models still have that.
4. Let's wordsmith the language strings.
   1. "Content" is a little vague for `fd.tc.body`. Maybe "Inline T&C"?
   2. `fd.tc.body.desc` should be a bit more helpful. Something like "Terms and Conditions for inline rendering. May contain HTML. If provided, will be used instead of URL." (We should also document that last bit in the [wiki](https://spaces.at.internet2.edu/display/COmanage/Registry+Terms+and+Conditions).)

boshrin:
I think we're almost there...

1. The `raw` layout is conceptually identical to the `widget` layout. I don't think it makes sense to reference the widget layout instead, but `raw.ctp` should be identical to `widget.ctp` (ie: it needs the standard file header, etc). We could also just symlink `raw.ctp` to `widget.ctp`.
2. We could add a field to track the original URL (eg `source_url`), but it could also just be pasted into the body itself so I'm not sure it's worth the effort.

I thought we solved validation localization somewhere, but I guess not. It's probably not worth revisiting before the framework migration.

muisit:
I updated the 'raw' layout. The widget layout is not present yet in this branch, it was added later on in develop. I could merge the branch with master though and then replace the raw layout with the widget layout. In a sense, the window that pops up with the T&C content can be considered a widget. But OTOH mixing this can potentially cause conflicts in case of future changes to the Dashboard and widget implementations.

boshrin:
Reverted since feature-3.2 is the wrong branch

PR #53, opened 2018-06-01
Platform admin permission for dashboard (CO-1625)

skoranda:


boshrin:
I'd actually already fixed this locally but hadn't committed it yet.

PR #67, opened 2018-10-03
Fix for CO1665

muisit:
We link the Pipeline created CoPerson and CoPersonRole to the petition after OrgIdentity creation by attached OIS plugins. 

This also introduces a 'link name' setting for Pipelines that allow linking the primary name specifically (which is otherwise explicitely not linked)

### DIFF
--- a/app/Config/Schema/schema.xml
+++ b/app/Config/Schema/schema.xml
@@ -485,6 +485,7 @@
     <field name="sync_on_update" type="L" />
     <field name="sync_on_delete" type="L" />
     <field name="create_role" type="L" />
+    <field name="link_name" type="L" />
     <field name="sync_cou_id" type="I">
       <constraint>REFERENCES cm_cous(id)</constraint>
     </field>
--- a/app/Controller/CoPetitionsController.php
+++ b/app/Controller/CoPetitionsController.php
@@ -995,6 +995,25 @@ class CoPetitionsController extends StandardController {
             catch(Exception $e) {
 
             }
+          } else {
+            // if a pipeline created a CoPerson, link that to the petition instead
+            $this->linkCoPersonFromPipeline($id, $newOrgId['OrgIdentitySourceRecord']['org_identity_id']);
+
+            // If a pipeline created a person and/or role, the status was set to 'Active'.
+            // Correct this now, to avoid a situation where the enrollment is cut-off for
+            // some reason, but the CoPerson and the CoPersonRole are set active without
+            // approval
+            $coPersonId = $this->CoPetition->field('enrollee_co_person_id', array('CoPetition.id' => $id));
+            if(!empty($coPersonId)) {
+              $this->CoPetition->EnrolleeCoPerson->id = $coPersonId;
+              $this->CoPetition->EnrolleeCoPerson->saveField('status', StatusEnum::Pending, array('CoPetition.id' => $id));
+            }
+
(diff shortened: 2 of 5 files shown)


boshrin:
I think 85d946456b may supercede this PR? Not sure if link_name is still required...

muisit:
I see my code also sets the status of the CoPerson and CoPersonRole after linking, because the pipeline made them active even though the status should be pending. I am not sure how this works out in this code. I think I remember having an active CoPerson even if my Enrollment was Denied. Or perhaps having an active CoPerson while the email confirmation was pending. 
It could be that the change to the CoPetition::saveAttributes method fixes this though. I am not sure when that method is called, the intricacies of Cake are still beyond me.

I did an additional call in the execute_selectOrgIdentityId, but I don't think that was a proper place.

As to link_name: yes we need this. If someone's name changes for any reason AND we pipelined this information from the OIS to the CoPerson, we DO want the name of the CoPerson to change (or at least, allow configuration of this feature). It seems rather silly to NOT pipeline the name as well, but I can imagine some administrators getting marginally confused when someones name changes. It is my opinion that this confusion is better in any respect than to force users to update their names manually (or even worse: have to ask admins to do this for them) for all COs they are a part of.

skoranda:
> As to link_name: yes we need this. If someone's name changes for any
> reason AND we pipelined this information from the OIS to the CoPerson,
> we DO want the name of the CoPerson to change (or at least, allow
> configuration of this feature). It seems rather silly to NOT pipeline
> the name as well, but I can imagine some administrators getting
> marginally confused when someones name changes. It is my opinion that
> this confusion is better in any respect than to force users to update
> their names manually (or even worse: have to ask admins to do this for
> them) for all COs they are a part of. 

In many deployments the system of record (SOR) that may be injecting
information through the OIS and a pipeline is of little consequence to
the user and his representation in the CO.

That is, in many deployments the CO cares little (if any) about the
details coming from the SOR (often a SAML IdP) and only care about the
details of the CoPerson record. Many science research projects operate
that way. It is the information in the CoPerson record that is important
to the user and the CO and it should not be changed just because the
upstream SOR makes a change.

An example would be using the CoPerson name information to generate
author lists for a scientific paper. A user would rightly be annoyed if
the name on a paper changed because some upstream SOR decided to make a
change.

So I agrue this flow of information and whether or not it is
automatically updated must be configurable, and the default should be
that it is not automatic. 

Thanks,

Scott K

muisit:
I disagree with your arguments, but if this feature were to be included, backwards compatibility would require it to be disabled by default in any case. It would be a further confusing aspect of the interface though.

skoranda:
> I disagree with your arguments

Then you are disagreeing with long term COmanage Registry deployments
and use cases in production for some science VOs.

I see how the use case you advocate is important for some deployments,
but that does not invalidate the needs of other deployers.

boshrin:
If a Pipeline cause an Active record to be generated before the completion of an enrollment, then that's a bug and we should open a JIRA (and possible a new PR if appropriate) for it.

The primary name discussion takes us to CO-1351, which is basically the configuration of rules to resolve the selection of single valued attributes. (If an OIS name is updated and a Pipeline is connected, the pipelined name attached to the CO Person will be updated. However whether or not that should be the Primary Name is a deployment specific concern.)

I'm going to close this PR and the above issues can be redirected to their appropriate channels.

PR #68, opened 2018-10-09
Fix to allow plugins to execute optional Enrollment steps

muisit:
This adds an else-block at the end of the tests for the current step. The first test looks to see if this is a 'curPlugin', the next test looks for RequiredEnum::Required steps. This leaves out the phase where we are not in a 'curPlugin' and the step is RequiredEnum::Optional (which would be the initial case for any optional step).

The effect of this can be easily tested: create an execute_plugin_start step in any enrollment plugin and have it output some debug text (or redirect to a view). This code is only executed if the enrollment requires the start phase (ie: if it has some introduction text configured). Without the introduction-text, the start phase is effectively skipped.

### DIFF
--- a/app/Controller/CoPetitionsController.php
+++ b/app/Controller/CoPetitionsController.php
@@ -741,6 +741,10 @@ class CoPetitionsController extends StandardController {
             $redirect['token'] = $token;
           }
           
+          if($step == "start" && !empty($this->request->params['named']['return'])) {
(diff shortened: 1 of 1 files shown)


boshrin:
af10bcf was addressed by a7d0937905 (for CO-1667, PR #69). Since I don't think I can pull the PR commits apart, 0650a4f is addressed by 3bfbdcc030.

PR #69, opened 2018-10-17
Quick fix for CO1667

muisit:
Dashes in the base64 encoded parameter are replaced by slashes (which would be interpreted as path separators). This is merely a suggestion, but as a workaround was required for us locally, we are going to apply this in the Surfnet fork.

Against all etiquette, this concept PR also includes a parameter change for history notifications in case of plugin exceptions. This ought to have been a separate PR.


boshrin:
I went with a slightly broader approach, substituting . for +, _ for /, and - for =. Also picked up the parameter correction.

PR #70, opened 2018-10-17
Easy fix for CO-1668

muisit:
When CoPetitionsController calls isAuthorize in the beforeFilter method, the cur_co object is not yet set (which is done by the parent::beforeFilter call). 
To retrieve a missing CO-id in the RoleComponent::calculateCMRoles method, this fix calls the AppController::parseCOID method, which returns a valid CO-id if one can be found. This same call is used as basis to populate the cur_co object in AppController::beforeFilter, but it avoids having to retrieve the whole object.

Worst case scenario would be publicly accessible pages, where parseCOID would be called twice, because an empty variable is returned.

### DIFF
--- a/app/Controller/Component/RoleComponent.php
+++ b/app/Controller/Component/RoleComponent.php
@@ -444,7 +444,9 @@ class RoleComponent extends Component {
     $controller = $this->_Collection->getController();
     
     $coId = $controller->cur_co['Co']['id'];
-    
+    if(empty($coId)) {
+      $coId = $controller->parseCOID();
+    }
     // Figure out the revelant CO Person ID for the current user and the current CO
     
     $CoPerson = ClassRegistry::init('CoPerson');


boshrin:
Shouldn't PHP throw a warning or error if cur_co isn't set? In other words, I would expect the patch to be

```
if(!empty($controller->cur_co['Co']['id'])) {
  $coId = $controller->cur_co['Co']['id'];
} else {
  $coId = $controller->parseCOID();
}
```

muisit:
PHP gives a warning for unset parameters, but I don't know about unset array elements. Could be that PHP silently accepts and returns null in such a case, possibly depending on runtime settings.

But your suggestion is far better by any comparison. Do you want me to adjust the PR, or can you just fix this straight in develop?

boshrin:
OK, fixed in 4ae44e2bb1.

PR #74, opened 2019-02-05
Fix LDAPProvisioner attribute options and no record in LDAP (CO-1709)

skoranda:


boshrin:
Pushed a stray file from the attribute options commit that is almost the same (but with slightly more error checking).

PR #75, opened 2019-04-03
Co 1725 plugins cannot send invites

muisit:
Adds a null attribute to explicitely set the controller, forcing the Router::url to disregard the current Plugin when generating a route.


boshrin:
I coincidentally fixed this in 2fa1d88cdb a few days ago, on hotfix-3.2.x. It'll be merged to develop when 3.2.2 is released.

PR #79, opened 2019-04-29
Co1612 provision services

muisit:
Implementation of basic Service provisioning, merged into the recent queued-job-changes for provisionable entities.


muisit:
Thanks for the comments. Yes, the Plugin/ProvisionerJob was already updated to allow provisioning of services already. It was one of the main issues while merging my local dev to upstream.
No, I do not have write access. However, I have trouble finding places in the documentation that need to be updated. The only thing I can find is the ProvisionerJob plugin page, which has a limitative list of record types it can provision. I assume adding CoService to that list is less trouble than giving me write access. But I'd be happy to edit that page.

tomzeller:
Looks like the co_provisioning_exports table references co_services before it (the co_services table) is created, resulting in errors when the database is created/setup for the first time.

boshrin:
Yes, the schema ordering bug is fixed in 46fa43c051.

PR #83, opened 2019-05-09
Do not provision CoPerson records for some status values (CO-1740)

skoranda:


boshrin:
Superceded by `67e0fe9307`.

PR #85, opened 2019-06-06
Added additional code to parse the CO id 

muisit:
The REST parameters for the autogenerate-identifiers API call pass a CoPerson ID. This is used to set the cur_co using a custom parseCOID method.


boshrin:
I can't reproduce the original issue described in CO-1754, perhaps there's some other context needed?

muisit:
Apparently the interface for identifier assignments has changed. They now go through a Job queue and the link to generate this Job queue includes the CO identifier, so it can create a Job specifically for that CO to regenerate identifiers. Seems fixed in a different way.

boshrin:
Yes, that's right. Identifier Assignments are now job based ... that was the missing context :)

PR #86, opened 2019-06-06
Adjusted pagination to use the correct page url

muisit:
instead of the window.location, which may be different for implicit URLs like index pages, we now pass the Paginator->url(), which is used in other places in the paginator output.


boshrin:
@spartanicus thought he had already fixed this issue and is trying to figure out what happened.

PR #89, opened 2019-07-04
Filtering in Relink OrgIdenity redirects to index (CO-1779)

ioigoume:
- Added in the form the link id and the coperson id
- In the controller i check if there is a link id available. If this is the case, i assemble the correct url and redirect


boshrin:
Fixed in 681dea2 (see CO-1779).

PR #90, opened 2019-09-25
Fixing redirect controller when confirming an OrgIdentity email

ioigoume:
The controller redirect attribute is wrong. It should be org_identities but it is org_identity_id.


ioigoume:
Closing since the fix got messed up after transfering

PR #96, opened 2019-10-26
Do not unsubscribe unmanaged addresses from lists (CO-1805)

skoranda:


boshrin:
See comments in [CO-1805](https://todos.internet2.edu/browse/CO-1805).

PR #98, opened 2019-11-04
Add ssl context options to HttpServer (CO-1812)

tomzeller:
Needs UI work to toggle hiding ssl options.


boshrin:
Minor change request: the main language strings should be English, not field labels. So, eg `'fd.server.http.ssl_allow_self_signed'` should be something like `Allow Self-Signed Certificates`, and then the `.desc` could be something like `Allow the server to present self-signed certificates when connecting`.

tomzeller:
Turns out that ssl_verify_host is not needed.

Adjusted field labels.

tomzeller:
Any chance this can be merged ?

PR #101, opened 2019-11-10
Fix the GithubProvisioner plugin

duncan-brown:
I tried to use the GitHub provisioning plugin with COmanage 3.2.2 installed using @skoranda's https://github.com/Internet2/comanage-registry-docker containers and ran into a bunch of issues.

It looks like the plugin is pretty old, and hasn't been updated to work with PHP 7.x, Cake 2.x, and other dependencies (in particular https://github.com/guzzle/guzzle version 6.x and https://github.com/KnpLabs/php-github-api version 2.x)

This pull request fixes these issues and restores the plugin to a working state. Details of the changes are in the commit messages. 

- Raises issues:
  - [Github provisioners should be able to share a client/secret.](https://todos.internet2.edu/browse/CO-1819)
  - [githubProvisioner Invitations Support,](https://todos.internet2.edu/browse/CO-1818) although @lppekows and I tested this and while the user can join the organization when they are in pending limbo, they are not added to the actual team. Since permissions are based on the team membership, I think things are OK here.
 - [More than one error in provisioning leads to a greyed-out continue button
](https://todos.internet2.edu/browse/CO-1821) which means that all users and groups must exist. The hack around this is
```diff
diff --git a/app/AvailablePlugin/GithubProvisioner/Model/CoGithubProvisionerTarget.php b/app/AvailablePlugin/GithubProvisioner/Model/CoGithubProvisionerTarget.php
index ac79e7ab..0b85ac36 100644
--- a/app/AvailablePlugin/GithubProvisioner/Model/CoGithubProvisionerTarget.php
+++ b/app/AvailablePlugin/GithubProvisioner/Model/CoGithubProvisionerTarget.php
@@ -201,6 +201,8 @@ class CoGithubProvisionerTarget extends CoProvisionerPluginTarget {
       }
       
       if(!$githubid) {
+       // FIXME patch to work around greyed out dialog box
+       return true;
         throw new RuntimeException(_txt('er.githubprovisioner.github_id') . ' (CoPerson ID ' . $provisioningData['CoPerson']['id'] . ')');
       }
     }
@@ -316,6 +318,8 @@ class CoGithubProvisionerTarget extends CoProvisionerPluginTarget {
     }
     
     if(!$teamid) {
+      // FIXME patch to work around greyed out dialog box
+      return true;
       throw new RuntimeException(_txt('er.githubprovisioner.team.notfound'));
     }

```

- Fixes issues:
  - [githubProvisioner Membership Support](https://todos.internet2.edu/browse/CO-942) since the GitHub API now supports sending invitations to users who are not members of an organization when they are added to a team.

- Unresolved issues:
  - [GitHub Provisioner Database Cache](https://todos.internet2.edu/browse/CO-945) since I still use a file-based cache, although it should be straightforward to implement a cache using e.g. Redis using the [updated client](https://github.com/KnpLabs/php-github-api#cache-usage)
  - [Provision SSH Keys to GitHub](https://todos.internet2.edu/browse/CO-944) for the reasons described in [CO-944](https://todos.internet2.edu/browse/CO-944) and [CO-943](https://todos.internet2.edu/browse/CO-943)


duncan-brown:
@boshrin this had no conflicts, but I rebased to 3.2.3 just to check.

boshrin:
Hi @duncan-brown ,

Some initial comments:

1. The PR should be submitted against `hotfix-3.2.x` instead of `master`. (You can see our full Branch Management policy [here](https://spaces.at.internet2.edu/display/COmanage/Branch+Management).) There's a slight chance there won't be a v3.2.4 release (depending on how long v.3.3.0 takes), in which case will just merge this directly into develop at that time.
2. I'm assuming FlySystem is required by GithubClient? (While it is license compatible, in general we try to avoid the additional maintenance of adding new external dependencies when we can avoid it.)
3. I'm not sure we need to commit `app/AvailablePlugin/GithubProvisioner/Vendor/cache`? It's adding a lot of noise so it's hard to review the rest of the PR.

duncan-brown:
@boshrin thanks for the feedback, here are the responses:

 1. Should be straightforward, I'll close this PR and open a new one against `hotfix-3.2.x`.

 2. and 3. Yes, both of these are required by https://github.com/KnpLabs/php-github-api We need some kind of caching so that we don't hit the GitHub API [rate limiter.](https://developer.github.com/v3/#rate-limiting) The old version of https://github.com/KnpLabs/php-github-api had a built in cache, but now it uses cache plugins. I used the simplest filesystem cache that I could find, but that pulls in `cache/filesystem-adapter` and `FlySystem`. The dependency tree is
```
cache/filesystem-adapter 1.0.0 A PSR-6 cache implementation using filesystem. This implementation supports tags
├──cache/adapter-common ^1.0
│  ├──cache/tag-interop ^1.0
│  │  ├──php ^5.5 || ^7.0
│  │  └──psr/cache ^1.0
│  │     └──php >=5.3.0
│  ├──php ^5.6 || ^7.0
│  ├──psr/cache ^1.0
│  │  └──php >=5.3.0
│  ├──psr/log ^1.0
│  │  └──php >=5.3.0
│  └──psr/simple-cache ^1.0
│     └──php >=5.3.0
├──league/flysystem ^1.0
│  ├──ext-fileinfo *
│  └──php >=5.5.9
├──php ^5.6 || ^7.0
├──psr/cache ^1.0
│  └──php >=5.3.0
└──psr/simple-cache ^1.0
   └──php >=5.3.0
```
I think I'm supposed to check in all of the code that composer pulls in the `Vendor/` directory, but let me know if I committed too much.

duncan-brown:
I noticed an edge case when testing: if an enrollment flow adds a user to a CO Group that exists on GitHub, they are added to the GitHub team with the same name while their enrollment is still pending. It doesn't happen with groups associate with COUs, just vanilla groups. This is likely an easy fix, so I will add a patch to this pull request when I figure it out.

duncan-brown:
Todos before merging:

 - [ ] Legalese in NOTICES.
 - [ ] Fix edge case with CO Groups.

duncan-brown:
The bug with the groups was a simple fix. The group provisioning function was lacking the active status check that the person provisioning path had. I also added a `trim()` as I had a user that included a space after entering his GitHub username. The GitHub API doesn't sanitize thus.

duncan-brown:
@boshrin I think it's good now, over to you for review.

PR #102, opened 2019-11-18
[CO-1824]Add search by email in CoPerson's REST API

ioigoume:
Refactored coperson Match endpoint implementation/logic. Add email in search criteria list.


boshrin:
Also we'll need to update [this page](https://spaces.at.internet2.edu/display/COmanage/CoPerson+API) to indicate the new parameter available as of v3.3.0.

PR #110, opened 2019-12-20
CO-1327_Login_and_Logout_actions_should_be_configurable

ioigoume:
Added configuration variable Logout.url.If defined the Service will:
- first clear the Session, local Logout
- redirect to the SSO service, if defined, for logout

If COmanage is used with shibd module for Apache2 then use the endpoint defined under the Session tag. The Logout.url needs to be the full URL and not just the path.


ioigoume:
@boshrin could you please review.

boshrin:
Minor item: The language keys should be `fd.ef.foo` for consistency with the existing keys. (They should all probably be renamed `fd.cmp` or something, but let's not do that now.) They should also be sorted alphabetically within the `lang.php` file.

I'm about to close feature-4.0, so when you update this PR, please switch it to develop.

ioigoume:
@boshrin made the changes

ioigoume:
Rebased on top of develop branch

PR #111, opened 2019-12-23
CO-1845_REST_API_username_is_a_two_word_prefix

ioigoume:
Transform CO name in a lower case letter word with no spaces

### DIFF
--- a/app/View/ApiUsers/fields.inc
+++ b/app/View/ApiUsers/fields.inc
@@ -43,6 +43,14 @@
 
   // Inject the CO ID
   print $this->Form->hidden('co_id', array('default' => $cur_co['Co']['id']));
+  // Alter the request data username field. We want to remove the co prefix so that we can present it properly. The correct format will be restructured on the beforeSave event
+  if (!empty($this->request->data["ApiUser"]["username"])) {
+    $usernameSplit = explode('.', $this->request->data["ApiUser"]["username"], 2);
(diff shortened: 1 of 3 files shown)


app/View/ApiUsers/fields.inc
```diff
@@ -52,17 +60,24 @@
...
     </div>
     <div class="field-info">
-      <?php print ($e
-        ? $this->Form->input('username', array('class' => 'focusFirst', 'default' => $cur_co['Co']['name'].'.'))
-        : filter_var($api_users[0]['ApiUser']['username'],FILTER_SANITIZE_SPECIAL_CHARS)); ?>
+      <?php
+        $prefix = 'co_' . $cur_co['Co']['id'] . '.';
+        print ($e ? $this->Form->input('username', array(
+                                                    'class' => 'focusFirst',
+                                                    'label' => $prefix,
```
spartanicus: The use of the $prefix as the label here is clever, but probably not best from an accessibility standpoint. We have a label in the field-name above, so rather than produce the prefex as one as well simply wrap it with a span and print it out directly, e.g. `print '<span class="api-username-prefix">' . $prefix . '</span>';`

ioigoume: I made the change.

spartanicus:
Looks good - I approve this request.

PR #121, opened 2020-03-13
CO-1880_COU_admin_can_not_edit_COU_admins_group

ioigoume:
COU:admins members and owners should be able to edit the group and add members

### DIFF
--- a/app/Controller/CoGroupsController.php
+++ b/app/Controller/CoGroupsController.php
@@ -376,46 +376,7 @@ class CoGroupsController extends StandardController {
     $self = false;
     
     if(!empty($roles['copersonid'])) {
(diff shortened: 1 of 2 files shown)


boshrin:
If I read this correctly, this PR would change the [documented behavior](https://spaces.at.internet2.edu/display/COmanage/Registry+Permissions) so that COU Admins could manage _any_ group. We don't appear to have written any notes, but my recollection from last week's dev call was that we did not agree to make this change, but that instead COU Admins could simply be designated as Owners of their COU Admins group, which would then allow them to manage the group. (This would be a deployment specific action and not a code change.)

ioigoume:
Hi Benn,
thank you for your feedback. I refactored the code and updated the PR accordingly.

Ioannis

boshrin:
Superseded by PR #125.

PR #130, opened 2020-04-14
Add filtering to "select member for group" view. CO-1881

spartanicus:
This commit provides the ability to filter by member, owner, first name, last name, email, identifier, and status. (CO-1881)


spartanicus:
Note that pagination conditions are dealt with directly in the select() method of the CoGroupMembersController. Unlike the index() method that separates these out (up in StandardController), there is no select() method in CoGroupMembersController's ancestors.

spartanicus:
The force push: I'd forgotten to add the search.inc file to the original commit.

boshrin:
It's not clear that `$managed` is a sufficient check in `isAuthorized()`.

PR #133, opened 2020-05-04
CO-951_Allow_Null_COU

ioigoume:
Added support for Empty COUs
Scenarios tested:
- Add empty named COU manually from inside CO Person's profile
- Choose an empty COU name when configuring an Enrollment flow
- Choose an empty COU name when collecting user's attributes

_Upgrade guide in COmanage wiki should be updated as well._

Should we:
- ~~Advise the user to hide the COU field if s/he has as a default COU name value the empty one and has the modifiable configuration disabled? **- Declined**~~
- enable copying the attributes of a non-empty COU to the empty COU during enrollment? **- Develop this as a Plugin(Enrollee plugin)**

### DIFF
--- a/app/Config/Schema/schema.xml
+++ b/app/Config/Schema/schema.xml
@@ -2555,6 +2555,7 @@
     <field name="disable_ois_sync" type="L" />
     <field name="group_validity_sync_window" type="I" />
     <field name="enable_normalization" type="L" />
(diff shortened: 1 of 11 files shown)


ioigoume:
Hi @boshrin,
i rebased and retested the code. Everything looks from my side.

PR #134, opened 2020-05-05
CO-1884 Delete nesting when Group is deleted

Jeanesily:
Change is on line 44.


Jeanesily:
@boshrin I re-committed this without the whitespace changes.  Thanks!

PR #135, opened 2020-05-07
CO-1748_Global_localization_variables

ioigoume:
Localization variables from COmanage CO given global scope


boshrin:
Need to decide if this is a breaking change or not.

ioigoume:
Rebased and moved on feature-4.0

boshrin:
As per today's call, for consistency with CO Themes (and other similar configurations), the merge order should be swapped so that the Platform localizations are merged first, and then (if set) the CO localizations will override.

PR #137, opened 2020-05-11
CO-1885 handle nested group membership on expunge

Jeanesily:
In testing, I found that the deleteDependent function (used by the CoPerson expunge) gathers up all CoGroupMember records, then deletes them in order.  The beforeDelete throws an error on trying to delete the Parent membership associated with a nested group membership, saying 'Record is already marked deleted.' 

That record was deleted in the afterDelete cleanup associated with the nested group membership's initial deletion - via the syncNestedMembership function (https://github.com/Internet2/comanage-registry/blob/develop/app/Model/CoGroupMember.php#L202).  

So!  To fix this (and also hopefully escape any other delete conflicts on expunge moving forward), let's check whether a record has already been cleaned up before trying to delete it.

### DIFF
--- a/app/Model/CoPerson.php
+++ b/app/Model/CoPerson.php
@@ -305,10 +305,15 @@ class CoPerson extends AppModel {
 				$records = $Model->find('all', array(
 					'conditions' => $conditions, 'fields' => $Model->primaryKey
 				));
-
-				if (!empty($records)) {
+        
+        if (!empty($records)) {
 					foreach ($records as $record) {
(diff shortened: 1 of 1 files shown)


Jeanesily:
A question for @boshrin whenever you have a chance - I noticed a lot of tabs still living in this file (which made the GitHub changes look a little funky).  I only cleaned up the lines I was directly working around.  Did you want me to fix any other tab issues in this file?  
Tabs

PR #138, opened 2020-05-14
CO-1912 Fix index error by on Expunge

Jeanesily:


Jeanesily:
@boshrin I can't assign you for review on this, looks like - but @ioigoume tested it out.  Thanks!

PR #140, opened 2020-06-03
CO-1896: Fix global search to focus on opening and submit on enter key

rmathis:
CO-1896: This fixes an issue where the form would not submit when a user hit the enter key after entering their search term. It also fixes the fact that the search input was not focused when the dropdown opened up.


spartanicus:
Thanks for the fix! This is approved for merge.

PR #144, opened 2020-06-08
Co 1459 Do Not Send Notifications to Expired Users

Jeanesily:
This beforeFind method is being called prior to running the query in CoNotification on line 609 (https://github.com/Internet2/comanage-registry/blob/develop/app/Model/CoNotification.php#L609).  

This particular elseif is being applied to the nested CoPerson contain (that is included within the CoGroupMember, which is contained within the CoGroup).  Here, a recursive call is done to modifyContain to walk through the CoPerson. 

Prior to the recursive call, the conditions for the CoPerson look like this: 
`'conditions' => array(
		'CoPerson.status' => 'A'
	)`

After the recursive call, the required new conditions of a null co_person_id and non deleted status are merged with the original Active conditions correctly: 
`'conditions' => array(
		'CoPerson.co_person_id' => null,
		(int) 0 => 'CoPerson.deleted IS NOT true',
		'CoPerson.status' => 'A'
	)`

Then, the conditions are changed again by being hard set to the additional required behaviors https://github.com/Internet2/comanage-registry/blob/develop/app/Model/Behavior/ChangelogBehavior.php#L642): 
`'conditions' => array(
		'CoPerson.co_person_id' => null,
		(int) 0 => 'CoPerson.deleted IS NOT true'
	)`

This is where the Active status condition is dropped.  I investigated instead merging any existing conditions with the new conditions, but you end up with an issue of duplicate conditions: 
`array(
	'CoPerson.co_person_id' => null,
	(int) 0 => 'CoPerson.deleted IS NOT true',
	'CoPerson.status' => 'A',
	(int) 1 => 'CoPerson.deleted IS NOT true'
)`

Because the recursive modifyContain call is correctly adding those conditions, let's remove the additional hard reset on those conditions.

### DIFF
--- a/app/Model/Behavior/ChangelogBehavior.php
+++ b/app/Model/Behavior/ChangelogBehavior.php
@@ -634,19 +634,8 @@ class ChangelogBehavior extends ModelBehavior {
               
               $ret[$k]['order'] = $v2;
             } elseif(is_array($v2)) {
-              //This recursive modifyContain will successfully add those conditions to a nested contain
-              $ret[$k][$k2] = $this->modifyContain($model->$k->$k2, $v2);
               
-              if(is_string($k2) && !is_integer($k2)) {
-                $cparentfk = Inflector::underscore($model->$k->$k2->name) . "_id";
-                
-                // This will then overwrite any existing conditions, and these conditions were already added in line
-                // 638 by the recursive modifyContain
-                // $ret[$k][$k2]['conditions'] = array(
-                //   $k2.'.'.$cparentfk => null,
-                //   $k2.'.deleted IS NOT true'
-                // );
-              }
+              $ret[$k][$k2] = $this->modifyContain($model->$k->$k2, $v2);
             } elseif(isset($model->$k->$k2)) {
               // Fifth example
               $m = $this->modifyContain($model->$k, array($k2 => $v2));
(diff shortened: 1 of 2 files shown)


boshrin:
There are two streams to this issue:

1. Generally, the concept of CO-1459 is that Notifications could be registered anywhere, and if the RecipientCoPersonID is no longer Active, they will still receive the Notification.
2. A potential bug in ChangelogBehavior, specifically in the context of Notifications where the recipient is a CO Group. In this case, the existing code tries to filter Active records, but perhaps doesn't.

For the general concern, the code currently performs no filtering. There are places where this is the desired behavior, specifically Enrollment Flows and Expiration Policies (and maybe Org Identity Sources). `CoNotification::register` will probably need a new parameter (eg `$notifyActiveOnly`) with existing calls being updated as appropriate for their context. Alternately, the invoking code could simply skip registering notifications if it already knows the recipient is not Active, but it's probably better to have the logic in one place.

For the specific concern, I'm not sure the proposed patch to ChangelogBehavior is correct.  There is a significant difference between

```
$ret[$k][$k2] = $this->modifyContain($model->$k->$k2, $v2);
```

and

```
$ret[$k][$k2]['conditions'] = array(...)
```

The former is adjusting the SQL query for the associated models (`$v2`), the latter is adjusting the SQL query for the primary model (`$k2`).

If Changelog isn't correctly filtering the records, one option would be to simplify the "clever contain" in `CoNotification::register`, and filter the desired records using `Hash::extract` or something after the find. (As ChangelogBehavior will be rewritten as part of framework migration, and is otherwise a bit fragile, I'm hesitant to change it if we can do something else.)

We can discuss on the next call if helpful.

PR #149, opened 2020-06-22
CO-1950_Default_value_textbox_preserves_value_among_attribute_selection

ioigoume:
Default value textbox should not preserve value while toggling on different attributes


ioigoume:
Rebased on top of hotfix-3.3.x

PR #151, opened 2020-06-24
Add OIDC sub, SAML pairwise-id subject-id (CO-1864)

skoranda:


boshrin:
Minor naming consistency nits...

For case, IdentifierEnum tends to follow the case of the thing its named after, such as "ePPN". So I would use `OIDCsub` rather than `OidcSub`.

Similarly, "ID" is sort of inherently redundant, so we only use it when its part of the existing abbreviation (eg "UID" or "ePTID"). So I would use `SAMLPairwise` and `SAMLSubject`.

PR #153, opened 2020-07-08
limit results for persons on group memberships

acousticrobot:
This PR has two solutions, one taken from a similar solution in the CoPeople controller for IFMC-1549 , and the other taken from existing code in the CoPeopleController: `$pagcond['conditions']['PrimaryName.primary_name'] = true;`


boshrin:
Minor nit... as for CO-1549 (PR #152), there should be a comment linking to the JIRA issue since we don't use group by too often and will probably lose track of why it's here.

PR #155, opened 2020-07-14
CO-1904: First field does not have input focus when manually adding new CO person role

kcontreras01:
https://todos.internet2.edu/browse/CO-1904

- Moved `focusFirst` to `Affiliation` field instead of `Title`


boshrin:
BTW this should be a PR against develop, not master (or hotfix-3.3.x if that's a thing). See https://spaces.at.internet2.edu/display/COmanage/Branch+Management

spartanicus:
Hi, Kiara! Sorry for the wait. 

This is good, but there's a nuance in the CO-1904 ticket: some Collaborations (COs) have COUs associated with them and some don't. Those with them will have the COUs appear as the first field in this form rather than Affiliation. The COU field is labelled "_COname_ COU". When it's exposed, it should be marked with the focusFirst class instead of Affiliation. You can try out the roles in different COs to see this - the COmanage CO is likely to have no COUs, while some of the others definitely will (try TestCO if you have it in your data).

You'll note the COU field is in the `<li>` element directly above the Affiliation field ...around line 176-211, and it has a test `if(!empty($permissions['cous']))`. You should be able to put the class on the COU form element and then use the test to decide if it should go on the affiliation below (by inverting the test to check for empty rather than !empty).

Thanks!

kcontreras01:
Thanks for looking at this @spartanicus. I tried checking flipping the test but no matter what I try it seems that 'affiliation' is the first item to be focused on. If you get the chance could you take a look at it? Thank you!

spartanicus:
Out of date - will close this for now and solve in another PR.

PR #165, opened 2020-08-26
CO-2023_Trying_to_access_non_existent_DOM_element

ioigoume:
Trying to access non existent CoEnrollmentFlowAuthzCoGroupId DOM element

### DIFF
--- a/app/View/CoSettings/fields.inc
+++ b/app/View/CoSettings/fields.inc
@@ -58,7 +58,7 @@
     if(authz == 'CG') {
       document.getElementById('sponsorgroupdiv').style.display = "block";
     } else {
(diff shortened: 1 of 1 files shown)


boshrin:
Based on the comment above, I don't think this is the right fix. Instead, the element ID should be corrected (to blank out the Sponsor CO Group ID when the sponsor mode is not CO Group).

boshrin (changes requested):
Based on the comment above, I don't think this is the right fix. Instead, the element ID should be corrected (to blank out the Sponsor CO Group ID when the sponsor mode is not CO Group).

ioigoume:
Change done.

PR #166, opened 2020-09-01
Configurable status when pipeline creates new CoPerson (CO-2021)

skoranda:


boshrin:
I would make the value optional. If not set, then the default Pipeline behavior executes. For now, the default behavior is to set to "Active" (which we should document), but we might change that behavior in a future release.

"If set, new CoPerson records will be created with this status"

PR #168, opened 2020-09-21
CO-2029_Enroller_plugin_start_hook_is_not_firing

ioigoume:
Let execute_start function handle the presence of an introductory text.

### DIFF
--- a/app/Model/CoEnrollmentFlow.php
+++ b/app/Model/CoEnrollmentFlow.php
@@ -490,13 +490,8 @@ class CoEnrollmentFlow extends AppModel {
     
     $ret = array();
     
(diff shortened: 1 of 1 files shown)


boshrin (changes requested):
I don't think this patch is correct. "Optional" denotes a setting where the core step does not run, but plugins may. This logic is executed in CoPetitionsController::dispatch. I don't see anything obvious that would be causing this to break for the start step, but that function is probably the place to start debugging.

ioigoume:
Closing as duplicate: https://github.com/Internet2/comanage-registry/commit/3bfbdcc030bac3ac4f4aa64469e676ac83433bf2

PR #170, opened 2020-09-24
CO-2027_Add_Support_for_HTML_and_BOTH_Email_formats

ioigoume:
Add Support for email text and html format

### DIFF
--- a/app/Config/Schema/schema.xml
+++ b/app/Config/Schema/schema.xml
@@ -703,6 +703,8 @@
     <field name="bcc" type="C" size="256" />
     <field name="message_subject" type="C" size="256" />
     <field name="message_body" type="X" size="4000" />
+    <field name="message_body_html" type="X" size="4000" />
+    <field name="format" type="C" size="4" />
     <field name="status" type="C" size="2" />
     <field name="created" type="T" />
     <field name="modified" type="T" />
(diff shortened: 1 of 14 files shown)


boshrin (changes requested):
1. The format enum values are not message contexts, and so should be in their own Enum, eg MessageFormatEnum. We can probably shorten the labels to something like `MessageFormatEnum::Plaintext`, `MessageFormatEnum::HTML`, and `MessageFormatEnum::PlaintextAndHTML`. ("Both" will be ambiguous if we ever add another format.)
1. I would tweak the language keys slightly: 'fd.mt.body.txt' = "Message Body (Plain Text) and 'fd.mt.body.html' = "Message Body (HTML)"
1. In `CoInvite::send()` you added a new parameter (`$format`) in the middle of the argument list, but it supports a default value. It should be placed at the end of the parameter list to minimize chances of breaking existing uses. (This is just a general best practice, I'm sure you already checked for existing uses.)
1. In `CoMessageTemplates/fields,inc`, you have a stray `console.log`. You also called the function `message_body_tongle`, but I think you meant `message_body_toggle`.

spartanicus:
Hi, Ioannis - the approach you've taken with the button (and spinner) looks great to me and serves a good purpose. I certainly sign off on this, and we can incorporate similar approaches elsewhere when needed.

boshrin:
Hey @ioigoume, I'm not sure if this was in the original PR, but in CoMessageTemplatesController you added the line

```
public $components = array('Security', 'RequestHandler');
```

Why was this necessary? The controller should pick these up via inheritance from `AppController`. (And by overriding here, other components added in AppController may not be loaded.)

Also, you have a couple of XXX notes in CoPetition. Did you mean to do something with those?

PR #171, opened 2020-10-05
CO-2034_Mis-calculation_of_Enrollment_Flow_Approvers_priviledges

ioigoume:
Calculation of Approver's permissions during an Enrolment Flow are buggy


ioigoume:
@boshrin concluded the changes discussed during the Dev call.

PR #173, opened 2020-10-20
CO-1475: add validation of hidden field when default value is missing

carsonreinke:
https://todos.internet2.edu/browse/CO-1475

### DIFF
--- a/app/Model/CoEnrollmentAttribute.php
+++ b/app/Model/CoEnrollmentAttribute.php
@@ -112,8 +112,42 @@ class CoEnrollmentAttribute extends AppModel {
       'rule' => array('numeric'),
       'required' => false,
       'allowEmpty' => true
+    ),
+    'hidden' => array(
+      'ruleName' => array(
+        'rule' => array('boolean'),
+        'required' => true // All attributes class have a hidden value sent even if they are not applicable
+      ),
+      'ruleName2' => array(
+        'rule' => array('validateHidden'),
+        'required' => false, 
+        'message' => 'A hidden field must have a default value'
+      )
     )
-  );  
+  );
+
+  /**
+   * Validate that the field is allowed to be hidden based on the attribute and default value
+   *
+   * @param  array Array of fields to validate
+   * @return boolean True if allowed, false if not
+   */
+  public function validateHidden($a) {
+    // Hidden attribute is only applicable to CO Person Role attributes (type 'r'), CO Person attributes (type 'g'),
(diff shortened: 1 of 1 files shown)


app/Model/CoEnrollmentAttribute.php
```diff
@@ -112,8 +112,42 @@ class CoEnrollmentAttribute extends AppModel {
...
+   * Validate that the field is allowed to be hidden based on the attribute and default value
+   *
+   * @param  array Array of fields to validate
+   * @return boolean True if allowed, false if not
+   */
+  public function validateHidden($a) {
+    // Hidden attribute is only applicable to CO Person Role attributes (type 'r'), CO Person attributes (type 'g'),
+    // Organizational Identity attributes (type 'o'), or Extended Attributes (type 'x')
+    $attrCode = explode(':', $this->data['CoEnrollmentAttribute']['attribute']);
+    if($attrCode[0] != 'r' && $attrCode[0] != 'g' && $attrCode[0] != 'o' && $attrCode[0] != 'x') {
```
Jeanesily, line 139: Nit picky - but would in_array work here instead of stringing several checks together?

carsonreinke: @Jeanesily yep, I saw this somewhere else, but the `in_array` thing is way more readable.

boshrin:
Some comments:

1. `ruleName` and `ruleName2` aren't all that descriptive.
1. `required` validation rules should be accompanied by `allowEmpty` rules (with an inverse value) due to technicalities in the framework. I'm not sure what it means that two rules have opposite required values... (`required` means "this key must be in the form data", `allowEmpty` means "the value for this key may be empty".)
1. Unlocalized strings should not be hardcoded in validation rules. (There are some legacy instances that haven't been clean up yet, but those shouldn't serve as examples.) Instead, the validation rule should return a localized text string on failure. See `AppModel::validateInput()` for an example.

boshrin:
Also, I'm not sure why this was flagged for v4.0.0 rather than v3.3.2, so this PR should probably be rebased against hotfix-3.3.x.

Jeanesily:
@boshrin Closing this out as I recreated/rebased it here: https://github.com/Internet2/comanage-registry/pull/182 (including the original commit: https://github.com/Internet2/comanage-registry/pull/182/commits/13adaba258dee8c74ad1141c391da751a7f94d20)

Jeanesily:
Oops, actually I can't close this @boshrin  if you could do that, thanks!

boshrin:
Closing per @Jeanesily

PR #174, opened 2020-11-03
 CO-929_Hierarchical_Representation_Of_Cous

lionick:
Add hierrachical representation of cous on my population list and CoPersonRole edit view dropdown list

### DIFF
--- a/app/Controller/CoPersonRolesController.php
+++ b/app/Controller/CoPersonRolesController.php
@@ -306,6 +306,22 @@ class CoPersonRolesController extends StandardController {
       $this->set('vv_extended_attributes', $this->extended_attributes);
       // (Dis)allow Empty COUs
       $this->set('vv_allow_empty_cou', $emptyCousPermit);
+
+      $roles = $this->Role->calculateCMRoles();
+      $this->set('vv_available_cous', array());
+      // Determine which COUs a person can manage.
+      if($roles['cmadmin'] || $roles['coadmin']) {
+        // Note that here we get id => name while in CoPeopleController we just
+        // get a list of names. This is to generate the pop-up on the edit form.
+        $cous = $this->CoPersonRole->Cou->find('threaded',array('conditions' => array("Cou.co_id" => $this->cur_co['Co']['id'])));
+        $childCous = array();
+        foreach($cous as $cou) {
+          $childCous = array_unique($childCous + $this->CoPersonRole->Cou->childCousById($cou['Cou']['id'], true, true));
+        }
+        $this->set('vv_available_cous', $childCous);
+      } elseif(!empty($roles['admincous'])) {
+        $this->set('vv_available_cous', $roles['admincous']);
+      }
     }
   }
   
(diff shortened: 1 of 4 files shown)


spartanicus:
This looks good to accept and works well. Approved for merge.

boshrin (commented):
I'm a little confused by the threaded find in CoPersonRolesController, since it appears to be using Model calls and then re-finding on top of that. Shouldn't this be using [Tree Behavior](https://book.cakephp.org/2/en/core-libraries/behaviors/tree.html), which is what the Cou.php does?

I'm also not understanding the new `$parent` parameter passed into `Cou::childCousById()`, especially since the passed in parameter might then immediately be masked by a local variable of the same name.

What is the old and new expected behavior here?

lionick:
Hi Benn, I'm glad you ask!
Before there was `$p['cous'] = $this->CoPersonRole->Cou->allCous($this->cur_co['Co']['id']);` which now is not sufficient as it returns all cous regardless level of cou and also returns only names . What we want now is to get the top level cous (not only theirs names but also ids). 
So firstly I thought to use **children** function which returns all top level cous BUT then I noticed that I cannot pass co_id so to have the right cous (pls correct me if i'm wrong).
So in order not to create another function at Cou.php I used **threaded** so I can have top level Cous.
If there is another sufficient way (e.g creating a new function) please share it.

Regarding the new argument at childCousById you are right is redundant, and forgotten after the refactoring. I will remove it and update the PR. Sorry for that.

lionick:
Hi Benn, I have updated the PR.

boshrin:
A couple of things.

First, the use of `$p['cous']` being set in `isAuthorized()` was confusing, because it is an old style of coding that we don't use anymore. I realize this was already there, but since you're working with this code, we should fix this to align with newer standards. Specifically, variables (like those used for select lists) should be set in `beforeRender()` and be prefixed `$vv_`. So the goal should be to add something like

```
$this->set('vv_available_cous', $this->CoPersonRole->Cou->childCousById(...));
```

to `CoPersonRole::beforeRender()` (and then remove the code that sets `$p['cous']` from `isAuthorized()`).

Second, I'm concerned about the changes to `Cou::childCousById()` since this function is also called from other places, and specifically `RoleComponent::couAdminFor()`. Have the changes been sufficiently tested to ensure `couAdminFor` (and the code that calls it) behaves as expected? It might be preferable to maintain the current behavior by default, but add a new flag to request the hierarchy in the name. eg:

```
public function childCousById($parentCouId, $includeParent=false, $includeHierarchy=false) {
  ...
}
```

lionick:
Hi Benn, I will do the changes regarding $p['cous'].

As for Cou::childCousById() yes i can add a new flag. 
RoleComponent::couAdminFor() is called in two places one is at calculateCMRoles() and that affects the left menu under People where we can see the hierarchal representation of the cou (e.g My vo.example.eu / vo.example-sub-1.eu / vo.example-sub-2.eu Population). 
The second place is at RoleComponent:isCouAdminForCoPerson() where it only cares for the cou_id not the cou name. So it doesn't affect the functionality in that place.

lionick:
Hi Benn, I've just updated the PR with the requested changes.

PR #178, opened 2020-11-26
CO-2054_UI/UX_CO_Person_canvas_refactor_field_presentation-action_list

ioigoume:
Add button action list for Names, Identifiers, mvpa.ctp, Groups.

### DIFF
--- a/app/View/CoPeople/fields.inc
+++ b/app/View/CoPeople/fields.inc
@@ -804,12 +804,30 @@
           <ul id="tabs-group" class="fields data-list">
             <?php
             if(isset($co_people[0]['CoGroupMember'])) {
+              // List of NON auto groups
+              $group_memberships = array();
               foreach($co_people[0]['CoGroupMember'] as $g) {
                 if($e && !$es && !$g['CoGroup']['auto']) {
                   if($g['member'] || $g['owner']) {
                     print '<li class="field-data-container centered">';
                     print '<div class="field-data">';
-                    print $this->Html->link($g['CoGroup']['name'], array('controller' => 'co_groups', 'action' => 'edit', $g['co_group_id'], 'co' => $cur_co['Co']['id']));
+                    if(strpos($g['CoGroup']['name'], 'CO:') !== false) {
+                      $gname = $this->Utils->systemGroupHFormat($g['CoGroup']['name'], $cur_co);
+                      print $this->Html->link(
+                        $gname['name'],
+                        array(
+                          'controller' => 'co_groups',
+                          'action' => 'edit',
+                          $g['co_group_id'],
+                          'co' => $cur_co['Co']['id'],
+                        ),
+                        array(
+                          'escape' => false,
+                        )
+                      );
+                    } else {
+                      print $this->Html->link($g['CoGroup']['name'], array('controller' => 'co_groups', 'action' => 'edit', $g['co_group_id'], 'co' => $cur_co['Co']['id']));
+                    }
                     print '</div>';
 
                     $badge_list = array();
@@ -926,37 +944,48 @@
                   }
                 } else {
                   $badge_list = array();
-                  print '<li class="field-data-container">';
-                  print '<div class="field-data field-data-alone">';
+                  $auto_group_format = $this->Utils->systemGroupHFormat($g['CoGroup']['name'], $cur_co);
                   if($g['member']) {
-                    print filter_var($g['CoGroup']['name'], FILTER_SANITIZE_SPECIAL_CHARS);
-                    if(!$g['CoGroup']['auto']) {
-                      $badge_list[] = array(
-                        'order' => BadgeOrderEnum::Type,
-                        'text' => _txt('fd.group.mem'),
-                        'color' => BadgeColorModeEnum::Gray,
-                      );
+                    // Do not present Auto Groups to non CO and Platform Administrators
+                    if(!$permissions['viewautogroups'] && $g['CoGroup']['auto']) {
+                      continue;
                     }
+                    $badge_list = array_merge($badge_list, $auto_group_format['badge']);
                   }
                   if($g['owner']) {
-                    print filter_var($g['CoGroup']['name'], FILTER_SANITIZE_SPECIAL_CHARS);
                     $badge_list[] = array(
                       'order' => BadgeOrderEnum::Owner,
                       'text' => _txt('fd.group.own'),
                       'color' => BadgeColorModeEnum::Blue,
                     );
                   }
-                  print '</div>';
-                  // Add Badge elements
-                  print '<div class="field-data data-label">';
-                  if(!empty($badge_list)) {
-                    print $this->element('badgeList', array('vv_badge_list' => $badge_list));
+                  if(!empty($group_memberships[$auto_group_format['key']])) {
+                    $badge_text = Hash::extract($group_memberships[$auto_group_format['key']]['badge'], '{n}.text');
+                    foreach($auto_group_format['badge'] as $bg) {
+                      if(!in_array($bg['text'], $badge_text)) {
+                        $group_memberships[$auto_group_format['key']]['badge'][] = $bg;
+                      }
+                    }
+                  } else {
+                    $group_memberships[$auto_group_format['key']] = $auto_group_format;
                   }
-                  print '</div>';
-                  print '</li>';
                 }
+              } // foreach($co_people[0]['CoGroupMember'] as $g)
+              // Print the Groups
+              foreach($group_memberships as $gr_membership) {
+                print '<li class="field-data-container">';
+                print '<div class="field-data field-data-alone">';
+                print $gr_membership['name'];
+                print '</div>';
+                // Add Badge elements
+                print '<div class="field-data data-label">';
+                if(!empty($gr_membership['badge'])) {
+                  print $this->element('badgeList', array('vv_badge_list' => $gr_membership['badge']));
+                }
+                print '</div>';
+                print '</li>';
               }
-            }
+            } // isset($co_people[0]['CoGroupMember'])
             ?>
           </ul><!-- tabs-group -->
         </li><!-- fields-group -->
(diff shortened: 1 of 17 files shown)


spartanicus:
This is great Ioannis! There are a couple of things that need to be fixed:

First - the caching on the menuAction element is currently causing each element to have the same ID, which breaks the drop-down menus when there are multiple elements in play. It may be best to remove all caching on the menuAction element (i.e. strip out the options arrays altogether). Because each element must include unique IDs, the element must be generated fresh each time. Alternately you could set a key on each call. Note the element caching information at https://book.cakephp.org/2/en/views.html#elements "If you render the same element more than once in a view and have caching enabled, be sure to set the ‘key’ parameter to a different name each time. This will prevent each successive call from overwriting the previous element() call’s cached result." If we do choose to cache, the cached elements need to be stored in a subfolder of tmp/cache/ and should be cleared when we call ./Console/cake cache. I suspect we can start without it.

Second - md5($vv_attr_value) will not guarantee unique IDs for the menu items because two $vv_attr_value variables can be identical (e.g. two name fields), and md5() will produce the same output for the same input. Instead, we could add a simple unique identifier at the top of the menuAction.ctp, e.g. $uid = uniqid(); and then change all four instances where you call  md5($vv_attr_value) just to $uid.

Third (and minor) - it would be great to remove the margin between the drop-down menu <li> elements (inherited from "ul.data-list li" - so that should be overridden on the drop-down elements).

I look forward to getting this merged - the menuAction element is a great improvement, and it will be nice to port this to all the views that need it!

spartanicus (changes requested):
See the main comment for details, but a) we need to remove caching on the menuAction element (or change the way it caches), b) use a different technique for producing unique IDs, and c) remove margin between the list elements in the drop-down.

ioigoume:
Hi @spartanicus ,
i addressed the requested changes and i am marking the PR as ready for a second review.

Thank you.

spartanicus (changes requested):
Hi, Ioannis - this is looking very good. We need some minor stylistic tweaks to ensure accessibility of the page. There are three things to improve (and one minor stylistic tweak):

1. The .highlight background color isn't accessible (i.e. the color contrast ratio doesn't pass WCAG 2.1 AA guidelines); to that end, let's keep this simple and drop a dashed outline on it in gray. Note too that .highlight is slightly too generic for my liking, so let's limit it to .data-list or at least to .field-data-container (co-base.css line 2035):

.field-data-container.highlight {
  border: 1px dashed #bbb;
}

2. The gear dropdown anchors must have text in them for screenreaders. The text could be "actions". We can hide the text with the .sr-only class when we use bootstrap. For now, we can create this class ourselves and just use the bootstrap style for "visually-hidden()" as our .sr-only class: https://github.com/twbs/bootstrap/blob/XhmikosR-patch-4/scss/mixins/_visually-hidden.scss (we can pull this out of our css when we move to Bootstrap).

3. We need to make sure that the .field-actions items focus visibly during keyboard navigation. Be sure that either ".field-actions .mdl-menu__item a:focus" turns on the background color or just allow the browser default behavior through (user-agent outlining).

4. Lastly - we should add some left-right padding (padding: 0 1em) to the .field-actions items -- the text "Authentication Events" is bumping the drop-down box edge. (co-base.css line 2266)

Again - this is looking very good!

ioigoume:
Hi @spartanicus ,
i made the changes derived from your review comments.

Ioannis

spartanicus (commented):
This looks great! I'm ready to approve it, but I note that there are still some sections on the canvas that should receive the new action drop-down treatment: Role Attributes and Organizational Identities - can we bring these in line too?

spartanicus (approved):
This is ready for merge - we can update the Role Attributes and Org Identities on a separate ticket.

spartanicus (dismissed):
This is ready to be merged. Thanks, Ioannis for the great work!

boshrin:
⚠️ Before merging let's see what conclusions come from Wednesday's Dev meeting (2/3).

---

Otherwise, my comments are mostly about the enums.

1. `BadgeColorModeEnum` seems like it should be defined functionally rather than by color. ie: In the code, I should say "this is a warning badge" or "this is an informational badge" rather than "this is a yellow badge".
2. There's an extra `o` in `MenuActionoOrder` (Action**o**).
3. I'm not sold on the `BadgeOrderEnum` or `MenuActionOrder` concepts. Historically, we've ordered items (1) alphabetically, (2) in the order they are listed in the code, or (3) by an `ordr` column. It would be nice if we could avoid a new mechanism.

spartanicus:
This is looking really good. There are some accessibility issues we need to address in the latest pass - primarily with text size and with contrast. I'll reach out directly.

spartanicus:
Hi, Ioannis - here are the basic items we need to address wrt accessibility:

**Size:**
- Avoid using the <small> tag. It is going to cause issues with the size of the fonts (and it's arguable that these are not the best semantic choice to wrap the badges in as well). Replace these with <span> where they are needed.
- The .badge style should be font-size: 0.95em; and font-weight: normal. This along with removal of the <small> tags will avoid creating badges that are too small. As we move to the version 5, we can discuss overall changes in font size as a group, as it is fairly small overall - but for now we need to avoid creating tiny text.

***Contrast:***
- We can use the .text-muted class, but we need to override its color to #555. This will allow the contrast ratio to be acceptable on backgrounds down to #ddd. We can use #666 on white and #eee, and that's also fine - but make sure anything on the #ddd background has a text color no lighter than #555. (I'll add some commentary on the style guide about this.)

***Other notes:***
- Please avoid the use of "&nbsp;" use normal white space and css for padding/margin where needed.
- Be judicious in the use of !important in the css - I'm not making a specific recommendation on this PR, but avoid it where you can.

spartanicus:
Thanks for the fixes Ioannis - this looks really good. My only remaining nit is really trivial: there is an extra <span> being rendered on the "Automatic" badge for groups that leaves it slightly misaligned (due to the subsequent nesting and flexbox). Can that be removed? 

image

app/View/CoPeople/fields.inc
```diff
@@ -556,73 +608,140 @@
...
+                $badge_list[] = array(
+                  'order' => BadgeOrderEnum::Source,
+                  'text' => filter_var($n['SourceName']['OrgIdentity']['OrgIdentitySourceRecord']['OrgIdentitySource']['description'],FILTER_SANITIZE_SPECIAL_CHARS),
+                  'color' => BadgeColorModeEnum::Gray,
+                );
+              }
+              // Name Type
+              $badge_list[] = array(
+                'order' => BadgeOrderEnum::Status,
+                'text' => filter_var($vv_cop_name_types[ $n['type'] ],FILTER_SANITIZE_SPECIAL_CHARS),
```
spartanicus, line 649: I think we'll want to continue using _txt('fd.name.primary_name') for the text of the badge here rather than the internal string reference ( $vv_cop_name_types[ $n['type'] ] ) - this will allow us to keep title case and i18n

ioigoume: Done

ioigoume:
Review fixes done.

spartanicus:
This is great work, Ioannis! From the UI's perspective, this is ready for merging.

spartanicus:
Thanks for the final updates, Ioannis. This is good for merging.

boshrin (changes requested):
This mostly looks good, but the enums are bothering me. Generally, we use enums to reflect coded strings stored in the database. The new enums defined here don't match that pattern.

I think a better approach would be to leverage UtilsHelper (which has too generic of a name and should be renamed to something more descriptive). The view element could then do something like

```
'order' => $Helper->getBadgeOrder('type');
```
or
```
$action_icon = $Helper->getMenuActionIcon('view');
```

I'm also still not sold on the enums being specified by color rather than function.

boshrin (commented):
I like the use of BadgeHelper.

Regarding the badge color, it seems strange to me to be specifying the _color_ in the code when the string that is returned is a _label_. For example,
```
'color' => $this->getBadgeColor('Blue')
```
sets color to `primary`, which isn't a color. This seems to get set back to an actual color in CSS, but then why did we convert it to a label in the first place? It seems to make more sense to use the functional label in the PHP code as well, so
```
'color' => $this->getBadgeColor('primary')
```
which would either return `blue` (if we want the color mapping done in PHP) or `primary` (if we want the color mapping done in CSS). For the latter, though, we don't really need a lookup table at all, we could just do
```
'color' => 'primary'
```
which then gets mapped to blue in CSS.

cc'ing @spartanicus for comments.

ioigoume:
@spartanicus , @boshrin ,
for the case of Badge Color naming i used the following mapper
```php
    $color_map = array(
      'Success'   => 'success',
      'Danger'    => 'danger',
      'Warning'   => 'warning',
      'Primary'   => 'primary',
      'Secondary' => 'secondary',
      'Light'     => 'light',
      'Info'      => 'info',
      'Dar'       => 'dark',
    );
```
This way we accomplish:

1. Decouple Bootstrap badge class names from COmanage code base
2. Keep the naming convention introduce by Bootstrat, thus making it easy to understand.
Could you please provide your feedback? Do you agree?

spartanicus:
I agree with this approach - I think it's certainly better to have functional names, and choosing to align with Bootstrap's naming conventions (with the difference of case to keep boundaries) is as sensible as anything since a) we need a reasonable taxonomy and b) we're using that library. I think if we can change "Dar" to "Dark" (for consistency) this is clear from my perspective.

ioigoume:
> I agree with this approach - I think it's certainly better to have functional names, and choosing to align with Bootstrap's naming conventions (with the difference of case to keep boundaries) is as sensible as anything since a) we need a reasonable taxonomy and b) we're using that library. I think if we can change "Dar" to "Dark" (for consistency) this is clear from my perspective.

If fixed the typo. Thank you for that out

spartanicus:
Thanks, Ioannis - this looks great to me. :-) 
@boshrin - this is ready to merge from the UI's perspective.

boshrin (approved):
Sounds like we've reached consensus, this pull request is \<suzeorman>**APPROVED**!\</suzeorman>

PR #185, opened 2020-12-11
CO-1882: Group management updates

spartanicus:
This PR holds early improvements in group management, specifically CO-1916 (Break Group Management Page Into Component Tabs) and CO-1925 (Default Group Filtering Should Exclude Automatic Groups). The latter includes a simple API for managing session variables via ajax calls (SessionXhrController.php).


spartanicus:
This PR should be reworked slightly to adopt the newly merged Bootstrap library (with MDL now deprecated). Also - the console.log() items should be removed from the code.

spartanicus:
Refactoring now done to adopt Bootstrap for this PR (and console.log() items have been removed).

spartanicus:
The feature-co1882 branch has been rebased against develop.

boshrin:
Next steps are to performance test people picker autocomplete box (can use QA), then ask for feedback from dev group.

spartanicus:
This has been rebased against the latest develop.

boshrin (commented):
I see both `$nest_contains` and `$select_contains` defined but not used anywhere?

spartanicus:
This has been rebased against the latest develop

spartanicus:
Rebased against the latest develop.

app/Controller/CoGroupMembersController.php
```diff
@@ -759,13 +830,53 @@ public function updateGroup() {
...
+      //$rid = $this->Session->read('Auth.User.co_person_id'); // used in setMembership()
+      try {
+        if($this->CoGroupMember->isMember($gid,$pid)) {
+          // CoPerson is already a member
+          $this->Flash->set(_txt('er.grm.already',array($plb,$gnm)), array('key' => 'error'));
+        } else {
+          // Add CoPerson to the group
+          $this->CoGroupMember->addByGroupName($pid,$gnm);
+          // XXX This is likely the better choice - needs RoleComponent (the first parameter)
+          // $this->CoGroupMember->setMembership(null,$gid,$pid,true,false,$rid);
```
boshrin, line 866: Should this be addressed or removed?

app/Controller/CoGroupsController.php
```diff
@@ -74,7 +74,7 @@ class CoGroupsController extends StandardController {
    */
 
   function beforeFilter() {
-    $this->redirectTab = 'group';
+    $this->redirectTab = 'group'; // XXX legacy? remove?
```
boshrin, line 77: Yes this is legacy and can be removed (and therefore the whole `beforeFilter()` override).

app/Controller/CoGroupsController.php
```diff
@@ -227,24 +227,40 @@ function checkWriteDependencies($reqdata, $curdata = null) {
...
-          $this->Flash->set(_txt('er.gr.init'), array('key' => 'information'));
-          return false;
+    if(!$this->request->is('restful')) {
+
+      if($this->action == 'add') {
+        $cos = $this->Session->read('Auth.User.cos');
+        $this->redirectTarget = array('action' => 'edit', $this->CoGroup->id);
+
+        // Member of current CO? (Platform admin wouldn't be)
+        if (isset($cos) && isset($cos[$this->cur_co['Co']['name']]['co_person_id'])) {
```
boshrin, line 237: A platform admin _could_ be a member of the current CO, does that matter?

Also, if you test `isset($foo['bar'])` you don't need to also test `isset($foo)`.

app/Controller/CoGroupsController.php
```diff
@@ -227,24 +227,40 @@ function checkWriteDependencies($reqdata, $curdata = null) {
...
         }
       }
+
+      if($this->action == 'edit') {
+        // return to the index view with default filters in place after edit
+        $this->redirectTarget = array(
+          'controller' => 'co_groups',
+          'action' => 'index',
+          'co' => filter_var($this->cur_co['Co']['id'],FILTER_SANITIZE_SPECIAL_CHARS),
+          $this->CoGroup->id,
```
boshrin, line 258: This is confusing and non-standard... why does and `index` view take a record id?

app/Controller/CoGroupsController.php
```diff
@@ -465,6 +555,19 @@ function isAuthorized() {
     // View an existing Group?
     $p['view'] = ($roles['cmadmin'] || $roles['coadmin'] || $managed);
     
+    // Viewing members, email lists, and nested groups is available to all group members.
+    // Access to specific actions is controlled in the view.
+    $p['members'] = $roles['cmadmin'] || $roles['coadmin'] || $managed || $member;
+    $p['email_lists'] = $roles['cmadmin'] || $roles['coadmin'] || $managed || $member;
+    $p['nest'] =  $roles['cmadmin'] || $roles['coadmin'] || $managed || $member;
```
boshrin, line 562: I'm not sure `|| $member` is appropriate here. The only permission that was previously intended was for a member to see their own membership (ie: the `co_group_member`) view, and maybe that wasn't even fully thought out.

app/Controller/CoGroupsController.php
```diff
@@ -681,6 +792,64 @@ public function paginationConditions() {
       $pagcond['conditions']['CoGroup.group_type'] = $searchterm;
     }
 
+    // Exclude admin groups
+    if(!empty($this->params['named']['search.noadmin'])) {
+      $pagcond['conditions']['CoGroup.group_type <>'] = 'A';
```
boshrin, line 797: Use enums, not values.

```suggestion
      $pagcond['conditions']['CoGroup.group_type <>'] = GroupEnum::Admins;
```

app/Controller/CoGroupsController.php
```diff
@@ -681,6 +792,64 @@ public function paginationConditions() {
...
+    // Filter by membership and ownership
+    // First get the CoPersonID to work against either the current user, or the user being acted on in select mode
+    $coPersonId = $this->Session->read('Auth.User.co_person_id');
+    if($this->action == 'select') {
+      $coPersonId = filter_var($this->request->params['named']['copersonid'], FILTER_SANITIZE_SPECIAL_CHARS);
+    }
+
+    // If both member and owner are selected, use a single join with "OR"
+    if(!empty($this->params['named']['search.member']) && !empty($this->params['named']['search.owner'])) {
+      $pagcond['joins'][] = array(
```
boshrin, line 809: Joins need to be tested against ChangelogBehavior. The easiest way to do this is to make changes in a record of the target table and then make sure you pull the newer values, not the older ones.

app/Lib/enum.php
```diff
@@ -499,6 +499,7 @@ class OrgIdentityStatusEnum
 class PeoplePickerModeEnum
 {
   const Sponsor   = 'S';
+  const CoPerson  = 'CP';
```
boshrin, line 502: Is this the right name? People Pickers are always picking CO People, the mode is intended to convey the context. So it should probably be something like `GroupMember`, or if there's nothing specific to groups then maybe `All` or `Active`.

app/Lib/lang.php
```diff
@@ -704,6 +704,12 @@
                                  NameEnum::Official => 'Official',
                                  NameEnum::Preferred => 'Preferred'),
   
+  // Nested Group member filters
+  'en.nested.filters' => array(
+    'e'      => 'Exclude',
```
boshrin, line 709: A proper enum should be defined in `enum.php`.

app/View/CoGroupMembers/select.ctp
```diff
@@ -79,98 +181,231 @@
...
+                  } else {
+                    print _txt('fd.group.own.only');
+                  }
+                } elseif($isMember) {
+                  print _txt('fd.group.mem');
+                }
+              }
+
+              // Warn if membership is invalid
+              // XXX Use badge after introduced by PR-178 / CO-2054
```
boshrin, line 301: PR-178 is merged?

app/Lib/enum.php
spartanicus: Good point - I think this mode can be used anywhere to grab a person; perhaps "All" is correct - I don't believe we're limiting the Group Member picker to Active only.

app/Controller/CoGroupMembersController.php
spartanicus: It should be removed.

app/Controller/CoGroupsController.php
spartanicus: Excellent - will do :-)

app/Controller/CoGroupsController.php
spartanicus: Shouldn't matter - agree that the comment is inaccurate. Will fix up comment and the test!

app/Controller/CoGroupsController.php
spartanicus: Yeah - this is worth a discussion. We can definitely scale this back - though it does beg the question of what a member sees when they visit a group they are in.

app/Lib/lang.php
spartanicus: I've updated the values, but the issues remains. I'll add these to enum.php as well.

app/View/CoGroupMembers/select.ctp
spartanicus: Yes! I'll get that in.

app/Controller/CoGroupsController.php
spartanicus: The record id was accidentally left in (bad copy paste, I suspect). Removed!

app/Controller/CoGroupsController.php
spartanicus: Fix committed

app/Lib/enum.php
spartanicus: Went with `const All       = 'AL';` for now

app/View/CoGroupMembers/select.ctp
spartanicus: Yes! I'll look at this.

app/Lib/lang.php
spartanicus: Added - we now have NestedEnum of which both lang.php and the CoGroupMembersController make use.

app/Controller/CoGroupMembersController.php
spartanicus: Update committed as of ee3fa84

app/Controller/CoGroupsController.php
spartanicus: Update committed as of ee3fa84

app/Controller/CoGroupsController.php
spartanicus: Update committed as of ee3fa84

app/Controller/CoGroupsController.php
spartanicus: Update committed as of ee3fa84

app/Lib/enum.php
spartanicus: Update committed as of ee3fa84

app/Lib/lang.php
spartanicus: Update committed as of ee3fa84

spartanicus:
First set of updates is complete, leaving three comments above still pending. I'm now looking at the joins (for changelog behavior) and the possible badging for invalid memberships. 

It's worth discussing what a regular member sees in a group (if anything) before I change the access to the group tabs for regular members. In develop, a regular member can see the Group in "view" mode - which provides access to the group properties, identifiers, nested groups, email lists, and the full members listing. This PR retains that behavior except for hiding the identifiers (which regular members don't need to see). 

The view of a Group is easily locked down from members, but how far do we lock it down? Do members see other members? Can members only see the group name in the list of their groups but not access the group view itself? Is that a Group RFE for setting on a case-by-case basis or CO setting? The current behavior is the same as develop, but we can certainly change it.

app/Controller/CoGroupsController.php
spartanicus: See [comment](https://github.com/Internet2/comanage-registry/pull/185#issuecomment-910264336) down below

app/Controller/CoGroupsController.php
spartanicus: This appears to be working; changes to member or owner status are immediately reflected in the pagination of the group (and group members) listing. (Tested: changes to member/owner status by an admin. The member immediately saw the change on reload/clicking a tab. Also tested updates to nested groups.)

The "Nested Groups" and "Email Lists" tabs both display the changelog at the bottom of their views, but these only reflect the Group properties changes. I suspect these should be removed so the context is clear. Do we need changelog behavior for these? The "Provisioned Services" has a changelog display too.

boshrin:
Based on today's dev call

1. "My Groups" and "My Memberships" should not show up in the menu for CMP admins who are not members of the CO.
2. "Groups" and "All Groups" need some clarification. We could have (eg) "Regular Groups", "System Groups", and "All Groups". Or "All Groups (unfiltered)". Or "Groups (default filter)".
3. "Provision" button should not render for COU admins, though they should be able to see the tab.

spartanicus:
"Regular Groups", "System Groups", and "All Groups" makes the most sense to me for item 1 - I think that distinguishes the difference between them fairly well and provides a nice shortcut to the "auto" groups. We'll start with that.

spartanicus:
This PR has been rebased against the latest develop.

spartanicus:
Notes 1 & 2 above are now fixed: "My Groups" and  "My Memberships" only appear for an active CoPerson in a CO, and the Group main menu links have been changed. 

This leaves three (known) items: read-only members list pagination (most important), restricting the "provision" button, and badges for invalid memberships.

spartanicus:
This has been rebased against the latest develop.

spartanicus:
Rebased again along with a fix that hides the Group "Provision" buttons from COU Admins.

PR #187, opened 2020-12-27
CO-2076_Existing_Message_Templates_Throws_Error_on_Notifications

ioigoume:
Added Upgrade script

### DIFF
--- a/app/Console/Command/UpgradeVersionShell.php
+++ b/app/Console/Command/UpgradeVersionShell.php
@@ -41,6 +41,7 @@ class UpgradeVersionShell extends AppShell {
                     'GrouperProvisioner.CoGrouperProvisionerTarget',
                     'HttpServer',
                     'Identifier',
+                    'CoMessageTemplate',
                     'SshKeyAuthenticator.SshKey',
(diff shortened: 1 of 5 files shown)


boshrin:
Is there a reason you need to unbind relations in `_ug400`? Is that just copy/paste from somewhere?

Also, this appears to include the updated text for the "Test Template" button. That should probably be a separate commit.

ioigoume:
> Is there a reason you need to unbind relations in `_ug400`? Is that just copy/paste from somewhere?
> 
> Also, this appears to include the updated text for the "Test Template" button. That should probably be a separate commit.

Regarding the unbinding of the relations i was wrong. According to the [documentation](https://book.cakephp.org/2/en/models/saving-your-data.html#model-saveall-array-data-null-array-options-array) `By default, updateAll() will automatically join any belongsTo association for databases that support joins. To prevent this, temporarily unbind the associations.`. So i fixed the code to unbind the belongsTo bindings instead of the hasMany.
I will make the Template Text a separate commit.

ioigoume:
@boshrin i applied the changes you proposed.

PR #188, opened 2020-12-31
CO-700_COU_Specific_Terms_and_Conditions_In_CO_Petition

lionick:


spartanicus:
This looks great from my side of things!

boshrin:
Minor comments...

1. In `Model/CoTermsAndConditions.php::getTermsAndConditionsByCouId`, start the `$args` block with `$args = array();`. While it isn't strictly needed here, it will potentially prevent copy and paste errors later.
2. Then later this line
```
if((!empty($cou_id) && ( $cou_id == $row['cou_id'] || $row['cou_id'] == NULL )) || empty($cou_id)) {
```
is a bit confusing as written. It would be more understandable as
```
if(empty($cou_id) || $row['cou_id'] == NULL || $cou_id == $row['cou_id']) {
```
3. Similarly, in `View/CoPetitions/petition-attributes.inc` the logic can be flipped so that
```
<?php if($vv_tandc_mode != TAndCEnrollmentModeEnum::ExplicitConsent
  && $vv_tandc_mode != TAndCEnrollmentModeEnum::ImpliedConsent): // action == view ?>
    return;
    <?php else: ?>
```
becomes
```
<?php if($vv_tandc_mode == TAndCEnrollmentModeEnum::ExplicitConsent ||
  && $vv_tandc_mode == TAndCEnrollmentModeEnum::ImpliedConsent): ?>
    return;
```
Note the removal of the `else` clause.

lionick:
Hi Benn, Thanks for these comments, were really helpful!
I have made the changes, and finished with a lot of tests. 
It seems to working fine to me.

PR #196, opened 2021-03-04
CO-1952_CO-2096_Auto_confirm_email_when_email_domain_matches_Identity_Providers_domain

ioigoume:
- Enrollment Flow Verified Mode.
- Verify CO Person Email if verified in OrgIdentity.

### DIFF
--- a/app/Config/Schema/schema.xml
+++ b/app/Config/Schema/schema.xml
@@ -1173,6 +1173,7 @@
       <notnull />
     </field>
     <field name="mail" type="C" size="256" />
+    <field name="skip_invite" type="L" />
     <field name="email_address_id" type="I">
       <constraint>REFERENCES cm_email_addresses(id)</constraint>
     </field>
--- a/app/Controller/CoPetitionsController.php
+++ b/app/Controller/CoPetitionsController.php
@@ -977,7 +977,9 @@ class CoPetitionsController extends StandardController {
           }
           catch(Exception $e) {
             $this->Flash->set($e->getMessage(), array('key' => 'error'));
-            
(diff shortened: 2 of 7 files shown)


boshrin:
`enum.php` defines `VerificationModeEnum::SelfReview`, but `CoEnrollmentFlow.php` uses `VerificationModeEnum::Verified`. Neither of these is exactly right, and maybe this should be a separate setting, or maybe it should always be enabled?

We should also document (in the code and/or JIRA) that `skip_invite` is a workaround for `sendConfirmation` needing to run… this should get cleaned up in the big enrollment flow rewrite (CO-1663).

ioigoume:
> `enum.php` defines `VerificationModeEnum::SelfReview`, but `CoEnrollmentFlow.php` uses `VerificationModeEnum::Verified`. Neither of these is exactly right, and maybe this should be a separate setting, or maybe it should always be enabled?
> 
> We should also document (in the code and/or JIRA) that `skip_invite` is a workaround for `sendConfirmation` needing to run… this should get cleaned up in the big enrollment flow rewrite (CO-1663).

As for the `VerificationModeEnum::SelfReview`, this is the correct and not the one in the CoEnrollmentFlow Model. I must have missed this when i renamed the parameter from `Verified` to `SelfReview`

boshrin:
Sorry for forgetting about this. I went back to my notes but couldn't figure out what I wanted to discuss. In rereading the PR, I think I just have one minor suggestion, which is that we change `SelfReview` to `SkipIfVerified`.

Then we just need to update the wiki with the new setting and the new database field.

PR #199, opened 2021-03-10
CO-2077_Don't_Allow_Illogical_Validity_Dates

ioigoume:
Do not permit illogical validity dates


boshrin:
I'd rather not add the `ValidationMessage` class, as it is not a pattern we use anywhere else. It's also not compatible with v5, which will use the standard `__()` localization methods. I _thought_ it was possible to just use `_txt()` in validation rules, but I can't see anywhere we've done this so maybe I'm misremembering.

The solution is  for `validateTimestampPrecedes` to return the error string (`_txt('er.validation.date')`) rather than `false`.

Some minor nitpicking:

1. `'er.validation.date'` should use double quotes within the message to avoid escaping, and for consistency with other entries (including, eg, `er.unknown` and `er.url`): `'"Valid From" date must precede "Valid Through" date'`
2. `validateTimestampPrecedes` doesn't seem like quite the right name, since in some cases the test is `<` and in others it is `>`. Maybe `validateTimestampRange`?

Finally, it looks like you're doing some frontend validation with `validate_date_input`? My concern here is simply that we don't really do much client side validation, so this is potentially scope creep. We should probably discuss with Arlen.

ioigoume:
@boshrin
review comments addressed.

ioigoume:
Pull Request refactored as discussed during the call.

boshrin:
Some minor nits while we're waiting for @spartanicus to review...

1. The name of the validation rule for `valid_through` in the model files should probably be `follows` instead of `precedes`.
2. There's an extra `;` in `AppModel.php` line 1559.
3. At `AppModel.php` line 1564, I don't think you need to check `is_bool()` since `===` requires a type match for `false`? Same at line 1579.

spartanicus:
This looks good to me.

ioigoume:
@boshrin ,
changes done.

-Ioannis

PR #200, opened 2021-03-13
CO-1835_registry_is_not_stripping_spaces_from_identifier

ioigoume:
Trim any leading or trailing whitespace from request data


boshrin:
I think the problem might be that [Normalization](https://spaces.at.internet2.edu/display/COmanage/Normalizing+Data) isn't running when the CO Petition is submitted (petitioner attributes). If this is the case, we'll also need to consider notifying existing deployments, since there may be side effects of data being submitted now being normalized when it wasn't before.

ioigoume:
Fixed

PR #203, opened 2021-03-16
CO-2102_Return_generic_validation_error_were_no_Message_key_is_defined

ioigoume:
Change validationErrors rule name to a default error message


boshrin:
Don't we need to do this in `StandardController::add` also?

boshrin:
From today's meeting, the new language key will be something like "Please recheck %1$s"

ioigoume:
@boshrin changed applied

tomzeller:
Looks fine for the couple of models I tried.

Working through this as an example use case for the test framework.

Will try to automate other models.

PR #204, opened 2021-03-20
CO-1885_cannot_expunge_user_if_member_of_nested_group

ioigoume:
CoGroupMember::isMember function must join CoGroups with CoGroupMembers

### DIFF
--- a/app/Model/CoGroupMember.php
+++ b/app/Model/CoGroupMember.php
@@ -28,7 +28,12 @@
 class CoGroupMember extends AppModel {
   // Define class name for cake
   public $name = "CoGroupMember";
-  
+
+  // All Nested Membership syncing
+  // CO-1885, using a class parameter as flag, since delete 
+  // function does not accept parameters like the save one
+  public $syncNested = true;
+
   // Current schema version for API
   public $version = "1.0";
   
@@ -834,7 +839,10 @@ class CoGroupMember extends AppModel {
     //     AND !$sourceMember
     //     AND member of all non-negated source groups for target
     //     AND not a member of any source group for target where Nesting/Negate = true
-    
+    if($this->syncNested === false) {
+      return;
+    }
+
     $shouldBe = false;      // Should $coPerson be a member of $targetGroup?
     $all = false;           // Is $targetGroup configured for nesting all mode?
     $negated = false;       // $coPersonId is ineligible for $targetGroup due to any negative membership
--- a/app/Model/CoPerson.php
+++ b/app/Model/CoPerson.php
@@ -319,6 +319,9 @@ class CoPerson extends AppModel {
               'conditions' => array('id' => $record[$Model->alias][$Model->primaryKey])
             ));
             if (isset(current($currentRecord)['deleted']) && current($currentRecord)['deleted'] != true) {
+              if(isset($Model->syncNested)) {
+                $Model->syncNested = false;
+              }
               $Model->delete($record[$Model->alias][$Model->primaryKey]);
             }
 					}


boshrin:
We stopped using `recursive` a while ago because it tends to pull in more data than necessary, and can sometimes do it in an inefficient way (depending on the model relations). We prefer more explicitly identifying the required associated data, eg with `contains`.

That said, it's not clear to me why this fixes the problem. What does this catch that wasn't being caught before?

ioigoume:
@boshrin 
the function `CoGroupMember::isMember()` needs the CoGroup model
```php
 public function isMember($coGroupId, $coPersonId) {
    $args = array();
    $args['conditions']['CoGroup.status'] = StatusEnum::Active;
```
and in order to achieve this it uses the default behavior which is `LEFT JOIN` all the `belongsTo` Models.
This is the query produced by the framework
```sql
SELECT COUNT(*) AS "count"
FROM "public"."cm_co_group_members" AS "CoGroupMember"
         LEFT JOIN "public"."cm_org_identities" AS "SourceOrgIdentity"
                   ON ("CoGroupMember"."source_org_identity_id" = "SourceOrgIdentity"."id")
         LEFT JOIN "public"."cm_co_groups" AS "CoGroup" ON ("CoGroupMember"."co_group_id" = "CoGroup"."id")
         LEFT JOIN "public"."cm_co_group_nestings" AS "CoGroupNesting"
                   ON ("CoGroupMember"."co_group_nesting_id" = "CoGroupNesting"."id")
         LEFT JOIN "public"."cm_co_people" AS "CoPerson" ON ("CoGroupMember"."co_person_id" = "CoPerson"."id")
         INNER JOIN "public"."cm_co_groups" AS "CoGroup"
                    ON ("CoGroup"."id" = "CoGroupMember"."co_group_id" AND "CoGroup"."co_group_id" IS NULL AND
                        "CoGroup"."deleted" IS NOT true)
WHERE "CoGroup"."status" = 'A'
  AND "CoGroupMember"."co_group_id" = 223
  AND "CoGroupMember"."co_person_id" = 1664
  AND (((("CoGroupMember"."valid_from" IS NULL) OR ("CoGroupMember"."valid_from" < '2021-03-20 16:49:41'))) AND
       ((("CoGroupMember"."valid_through" IS NULL) OR ("CoGroupMember"."valid_through" > '2021-03-20 16:49:41'))))
  AND "CoGroupMember"."co_group_member_id" IS NULL
  AND "CoGroupMember"."deleted" IS NOT true
```
because it uses the `$recursive = 1` found in `Model.php`.
As soon as we disable recursive in `CoPerson::deleteDependent()`
```php
			if ($data['foreignKey'] === false && $data['conditions'] && in_array($this->name, $Model->getAssociated('belongsTo'))) {
				$Model->recursive = 0;
				$conditions = array($this->escapeField(null, $this->name) => $id);
			} else {
				$Model->recursive = -1;
				$conditions = array($Model->escapeField($data['foreignKey']) => $id);
				if ($data['conditions']) {
					$conditions = array_merge((array)$data['conditions'], $conditions);
				}
			}
```
the query becomes
```sql
SELECT COUNT(*) AS "count"
FROM "public"."cm_co_group_members" AS "CoGroupMember"
WHERE "CoGroup"."status" = 'A'
  AND "CoGroupMember"."co_group_id" = 92
  AND "CoGroupMember"."co_person_id" = 1660
  AND (((("CoGroupMember"."valid_from" IS NULL) OR ("CoGroupMember"."valid_from" < '2021-03-20 15:23:30'))) AND
       ((("CoGroupMember"."valid_through" IS NULL) OR ("CoGroupMember"."valid_through" > '2021-03-20 15:23:30'))))
  AND "CoGroupMember"."co_group_member_id" IS NULL
  AND "CoGroupMember"."deleted" IS NOT true
```
and fails.

boshrin (changes requested):
As a minor nit, there should be a comment in the code indicating we're using `$syncNested` because `delete()` doesn't support passing of options. ie: What we really want is `delete($record, array('syncNested' => false))` but Cake 2 doesn't support that.

PR #205, opened 2021-03-23
CO-1630_No_Enrollment_Attribute_Default_Values_When_Using_Attribute_Enumerations

ioigoume:
vv_enums handle as json object.
Fetch attribute maps from DB if not enum available.

### DIFF
--- a/app/Model/AppModel.php
+++ b/app/Model/AppModel.php
@@ -1625,10 +1625,13 @@ class AppModel extends Model {
         // translated names. Note as of v2.0.0 there may not be "translated"
         // names (ie: for attribute enumerations), in which case we just want
         // the original string.
-        
+
         foreach($this->validate[$field]['content']['rule'][1] as $key) {
           if(isset($this->cm_enum_txt[$field])) {
             $ret[$key] = _txt($this->cm_enum_txt[$field], NULL, $key);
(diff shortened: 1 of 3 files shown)


spartanicus:
I see the Attribute Enumerations and can define a default value for an Enrollment Attribute of "Department (Organizational Identity)" - if this ticket aimed only to correct that, it appears to be working. However, what I don't see is the list of attributes (or the default value if set) in the enrollment flow. Is CO-1630 concerned only with the first issue or is it also concerned with the second?

ioigoume:
@spartanicus could you please review. I rebased and tested on top of the develop branch

spartanicus:
Hi, Ioannis - though I see the attribute enumeration in the Enrollment Flow configuration, I'm still only seeing a text input when the Enrollment Flow is run: 

`<div class="form-group">`
`      <label for="EnrolleeOrgIdentityOu">Org Dept</label>`
`      <input name="data[EnrolleeOrgIdentity][ou]" class="form-control" maxlength="128" type="text" id="EnrolleeOrgIdentityOu">`
`</div>`

I've got no errors behind the scenes and the cache is clear.

boshrin:
Not sure I'm understanding the need for changes to AppModel.php...

PR #206, opened 2021-03-25
CO-1965_default_ordering_for_identifier_assignments

ioigoume:
Enable default ordering for identifier assignments


boshrin:
Following the "thin controller, thick model" approach, this logic belongs in the model, not the controller. See `Model/CoProvisioningTarget.php` for an example.

ioigoume:
@boshrin 
Done

PR #207, opened 2021-04-02
Hotfix for CO-700 (when no T&C is present on an enrollment flow)

lionick:
Related to [PR-188](https://github.com/Internet2/comanage-registry/pull/188)
Fix a bug when no T&C is present on an enrollment flow

### DIFF
--- a/app/Model/CoPetition.php
+++ b/app/Model/CoPetition.php
@@ -1984,7 +1984,7 @@ class CoPetition extends AppModel {
     // check value of first row if it is bigger than 0
     // 0 means that we will not take it into account
     // CoTermsAndConditions should be always present at form even if it does not have a value for security reasons
-    if(!empty($requestData['CoTermsAndConditions']) && (empty($requestData['CoTermsAndConditions'][0]) || $requestData['CoTermsAndConditions'][0] != 0)) {
+    if(!empty($requestData['CoTermsAndConditions']) && (!isset($requestData['CoTermsAndConditions'][0]) || $requestData['CoTermsAndConditions'][0] != 0)) {
       $tAndCMode = $this->CoEnrollmentFlow->field(
         't_and_c_mode',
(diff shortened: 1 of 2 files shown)


spartanicus:
Hi, Nick - this does fix the fundamental issue that blocks an enrollment flow in a CO without Terms and Conditions. We note that the .modelbox div still appears:
image
which can be fixed with slightly more specific selectors for that (in this case "display: grid" is overriding the .hidden class).

What is likely better than hiding the div with CSS is to test in the view if Terms and Conditions are present at all, and only render the dummy box (for JavaScript construction) if they are. Is that achievable?

lionick:
Hi Arlen, this is something we have discussed together and with Benn. The dummy box is needed, in case we have a modifiable membership and somenone can change COU at the enrollemnt flow. Because some COUs may not have T&C and others may have. CakePHP forms want to know all the fields when the page renders so we can't put the field for T&C adhoc, it must be there the moment the form loads. So we chose to have a dummy field instead of disabling security of the form. I was pretty sure that when I made the previous PR that .modelbox div didnt appear when was hidden (but it seems to be there now :P).  
I am not a CSS expert but i think in that case .hidden class loses its usability.  Maybe it would be better to add a specific css rule for that case?

spartanicus:
Thanks, Nick - that makes sense (we do similar things elsewhere), and we can definitely just hide it with the following in co-responsive.css at around line 391 under the section  "/* PETITION ATTRIBUTES */" right below where display: grid is declared:

`body.petitionerAttributes .modelbox.hidden {
	display: none;
}`

lionick:
Thanks Arlen, I have updated the PR and tested it ( It took me some time to bring the new updates from develop branch and configure it).

spartanicus:
Looks great! Thanks, Nick :-)
I'll merge this now.

PR #210, opened 2021-04-12
CO-1718_deleting_a_CO_times_out

ioigoume:
Make Delete CO a job shell


skoranda:
There are some typos in the Lib/lang.php additions.

skoranda:
This commit adds a new behavior for models. I would like to see that become its own commit, and then have the commits that leverage it to fix CO-1718 come after that.

boshrin:
I see `CoSettingsController::job()` registers the GarbageCollector Job, but it looks like the admin has to manually click a button to get the GarbageCollector scheduled? Can the GarbargeCollector be initially scheduled in `SetupShell` (and `UpgradeVersionShell` for existing installations when they upgrade to 4.0.0)?

boshrin:
1. `CoSetting->_ug400_1` and `CoSetting->_ug400_2` should be merged into `CoSetting->_ug400`, which can implement both steps.
2. Regarding Job queuing via the UI (`$vv_jobs_queued`), I'd rather centralize all Job UI work to the CoJobs interface (see also [CO-1309](https://todos.internet2.edu/browse/CO-1309)), but maybe we should talk about this on the next dev call.

PR #219, opened 2021-05-04
CO-2144_Implement_filter_banner_for_Jobs_Index_View

ioigoume:
Add filter banner to CO Job index view


boshrin:
See also CO-2160 for potential generalization.

PR #222, opened 2021-05-12
Update petition-attributes.inc

ioigoume:
Make conclusion text occupy the entire row and not just one column


spartanicus:
Looks good. Merging.

PR #223, opened 2021-05-18
CO-2129 Fix layout for default Invite form

spartanicus:
This change separates out the default "Invite" form from the Person Canvas display on CoPeople/fields.inc. The invite form has been restructured similarly to other enrollment flows to fix the layout and make it consistent with our other forms.

### DIFF
--- a/app/View/CoPeople/fields.inc
+++ b/app/View/CoPeople/fields.inc
@@ -401,309 +401,189 @@
   });
 </script>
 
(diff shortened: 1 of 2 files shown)


spartanicus:
Hide white space changes to make comparison of the fields.inc files easier. The code was re-nested because there were multiple instances testing for action=invite that were refactored into one global test (the invite form and the person canvas are mutually exclusive).

PR #229, opened 2021-06-06
CO-834_Filter_CO_Petitions_By_COU

ioigoume:
Refactor CO Petition Index View Filtering Block


boshrin:
As per today's discussion @ioigoume will add shortcuts to View Pending Confirmation/Approval to the buttons.inc for the petitions index as an interim workaround for the change to the default behavior (PA instead of PA+PC by default) then ok to merge.

PR #230, opened 2021-06-09
CO-1923_Redirect_to_Edit_after_Add_Group

spartanicus:
Redirect to the Group Edit action after adding a group. ﻿﻿﻿This approach allows us to send arbitrary redirects from any controller.


spartanicus:
Something squirrely rearranged the order of the source code (likely a new setting in my IDE). Only about four lines should be changed in this pull request, and instead this has a bunch of order changes. I'm going to close the PR and delete the branch and try this again with a clean start.

PR #231, opened 2021-06-09
CO-1923_Redirect_to_Edit_after_Group_Add

spartanicus:
Redirect to edit group after adding a group. This approach allows for arbitrary redirects to be sent from any controller.


spartanicus:
I've modified vv_redirect (view variable) to cv_redirect (controller variable) to better reflect where/how this variable is to be used.

spartanicus:
Going to reset this pull request completely.

PR #232, opened 2021-06-15
CO-2166_validateTimestampRange_Fails_on_One_of_valid_from_and_valid_through

ioigoume:
valid_from valid_through could be missing in the case of OrgIdentitySource plugins


boshrin:
Oh here's another edge case to think about, though I'm not sure if we have any code that currently works like this: What if I do `$OrgIdentity->saveField('valid_through');`? This ordinarily will just provide `valid_through` to the callback, but if `valid_from` was already set we'd need to pull it from the database and compare...

PR #233, opened 2021-06-16
CO-1923_Redirect_to_Edit_after_Group_Add

spartanicus:
Redirect to edit group after adding a group. This approach allows for arbitrary redirects to be sent from any controller.


boshrin:
@spartanicus to reimplement as a StandardController variable rather than use viewVars.

PR #236, opened 2021-07-08
CO-2160_Some_CoPerson_index_filter_by_email_result_in_500

ioigoume:
Append op:search at the end of the url


spartanicus:
This is a good solution until the more general approach is taken.

PR #240, opened 2021-07-17
CO-2172_Set_App.base_location

ioigoume:
Introduce CMP App.base configuration


boshrin:
A couple of minor comments.

1. The login code in `UsersController.php` should call `getAppBase()`, and then only save the webroot if it differs. You should also be able to use `CmpEnrollmentConfiguration::findDefault` to pull the CMP config.
2. You should also be able to use `RoleComponent` to determine if the current user is an admin. (Membership in the COmanage CO is not strictly sufficient.) Something like `$this->Role->isCoAdmin($this->Session->read('Auth.User.co_person_id'), 1);` or `$this->Role->identifierIsCmpAdmin($REMOTE_USER);`
3. It occurs to me that if a platform admin logs in, changes the app base, and then logs in again, we'll just override whatever the platform admin just did. So we probably shouldn't even offer this (contrary to my previous suggestion). Let's just make `app_base` read only in `fields.inc` (we probably need to be careful in how that gets emitted so the form still saves correctly when other changes are made), so admins can at least inspect the value in case something weird is happening.

PR #242, opened 2021-07-28
CO-2179_Refactor_My_Population_filtering_block

ioigoume:
User search.ctp element in My Population Index View


app/View/Elements/alphabetSearch.ctp
```diff
@@ -0,0 +1,87 @@
...
+ * @package       registry
+ * @since         COmanage Registry v4.0.0
+ * @license       Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ */
+
+// Globals
+global $cm_lang, $cm_texts;
+// Get a pointer to our model
+$model = $this->name;
+$controller = Inflector::tableize($model);
```
spartanicus, line 32: tableize() adds an "s" to our model (in this case "co_peoples"). I suspect what we want is:
$controller = Inflector::underscore($model);

spartanicus (changes requested):
This looks really good - but we need to fix the Inflector::tableize() call in alphabetSearch.ctp, before merging.

app/View/Elements/alphabetSearch.ctp
ioigoume: Done

PR #247, opened 2021-08-14
Use Application Preferences for Main Menu state handling (CO-1928)

spartanicus:
Extend Application Preferences for use in all views. This provides a new view variable for accessing a user's application preferences and makes use of them to set the state of the main menu drawer. The legacy use of Cookies for this purpose is deprecated. A JavaScript setter function is added for AJAX PUT of Application Variables.

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -1031,7 +1031,7 @@ class AppController extends Controller {
 
   /**
    * Get a user's Application Preferences
-   * - postcondition: Application Preferences variable set if Auth.User.co_person_id exists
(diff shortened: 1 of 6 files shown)


boshrin (changes requested):
Your `if(!empty($prefs))` test on line 1052 will always return true since you define `$prefs` right above it with some content. You instead want to test `if(!empty($prefs['preferences']))`.

You could also simplify `retrieveAll` to `return $this->find('all', $args);`. You could use `find('list')` instead to return the associative array that you are manually building. (You already have the correct `fields` definition.)

PR #248, opened 2021-08-15
CO-2183_CoExpirationPolicy_Model_does_not_implement_Changelog_behavior

ioigoume:
CoExpirationPolicy model should implement changelog behavior


boshrin:
In `schema.xml` by convention the changelog foreign key (`co_expiration_policy_id`) should be defined after `modified`, with the other changelog metadata. Also, the `actor_identifier` field appears to be missing.

PR #250, opened 2021-08-19
CO-2180_Localizations_can_not_be_used_by_Shell_Jobs

ioigoume:
- Move localization loading into CoLocalization Model.
- Jobshell loads localization's

### DIFF
--- a/app/Console/Command/JobShell.php
+++ b/app/Console/Command/JobShell.php
@@ -28,6 +28,7 @@
 class JobShell extends AppShell {
   var $uses = array('Co',
                     'CoJob',
+                    'CoLocalization',
                     'CmpEnrollmentConfiguration',
                     'Lock');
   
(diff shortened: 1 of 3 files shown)


boshrin:
In `JobShell.php`, you can add `CoLocalization` to `$uses` and then you don't have to explicitly init it.

Similarly, in AppController, you can reference `$this->Co->CoLocalization`. (I'm not sure why we previously explicitly loaded `CoLocalization`, it shouldn't have been necessary.)

Also, don't we always want to load platform localizations? Per the [documentation](https://spaces.at.internet2.edu/display/COmanage/Registry+Localization), Platform level localizations (ie: defined in the COmanage CO) apply everywhere, but can be overridden by CO specific text. In other words, `CoLocalization::load($coid)` should always load Platform Localizations, then load the Localizations for `$coid` on top of them.

BTW, I just (yesterday) defined a constant `DEF_COMANAGE_CO_NAME`, which should be used instead of hardcoding `'COmanage'` as the `Co.name` value.

PR #251, opened 2021-08-24
CO-2017_Normalization_vs_Attribute_Enumerations

ioigoume:
- Render Dictionary Management action link only for Standard dictionaries
- AttributeEnumeration view `allow other values` field refactor


boshrin (changes requested):
1. `doNormalize` should just return `$value` (instead of `true`) if it doesn't do anything. This will eliminate the need to test the result when it is called.
2. Also in `doNormalize`, you don't need to test `empty($attrEnum)` since your more specific `empty` in the same clause implies a test on `$attrEnum` first.
3. Your comment says we run normalization if the configuration allows other values, but there's no logic for that -- and there probably shouldn't be. Maybe just remove the comment?
4. Maybe the logic in `doNormalize` should just be moved into `AppModel::normalizeEnumeration`. I realize you're following the `validateEnumeration` pattern, but the actual normalization work applies to everything _except_ AttributeEnumerations.

ioigoume:
>     1. `doNormalize` should just return `$value` (instead of `true`) if it doesn't do anything. This will eliminate the need to test the result when it is called.
> 
>     2. Also in `doNormalize`, you don't need to test `empty($attrEnum)` since your more specific `empty` in the same clause implies a test on `$attrEnum` first.
> 
>     3. Your comment says we run normalization if the configuration allows other values, but there's no logic for that -- and there probably shouldn't be. Maybe just remove the comment?
> 
>     4. Maybe the logic in `doNormalize` should just be moved into `AppModel::normalizeEnumeration`. I realize you're following the `validateEnumeration` pattern, but the actual normalization work applies to everything _except_ AttributeEnumerations.

@boshrin Addressed all the comments from above.

app/Model/AppModel.php
```diff
@@ -1533,6 +1533,48 @@ public function validateEnumeration($coId, $attribute, $value) {
...
+    $args['conditions']['AttributeEnumeration.attribute'] = $attribute;
+    $args['conditions']['AttributeEnumeration.status'] = SuspendableStatusEnum::Active;
+    $args['contain'] = false;
+
+    $AttributeEnumeration = ClassRegistry::init('AttributeEnumeration');
+    $attrEnum = $AttributeEnumeration->find('first', $args);
+
+    if(empty($attrEnum['AttributeEnumeration']['dictionary_id'])) {
+
+      $this->Behaviors->load('Normalization');
```
boshrin, line 1561: Sorry, for #3 I just meant to remove the part of the comment about "other values". The rest of the comment was useful :)

```suggestion
    if(empty($attrEnum['AttributeEnumeration']['dictionary_id'])) {
      // If there is no dictionary attached to the Attribute Enumeration
      // configuration then load the normalizer
      $this->Behaviors->load('Normalization');
```

ioigoume: Done

app/View/Dictionaries/buttons.inc
```diff
@@ -37,7 +37,10 @@ if(!$e && !$permissions['view'])
 
 if($e) {
   // Manage Dictionary Entries button
-  if($this->action == "edit" && $permissions['edit']){
+  // We only allow Dictionary Entries to be managed for Standard Dictionaries.
+  if($this->action == "edit"
+     && $permissions['edit']
+     && $dictionaries[0]['Dictionary']['id'] === DictionaryModeEnum::Standard){
```
boshrin, line 43: Sorry I just noticed this, I think you want to check `mode` here?

```suggestion
     && $dictionaries[0]['Dictionary']['mode'] === DictionaryModeEnum::Standard){
```

ioigoume: Done

PR #253, opened 2021-08-29
CO-2190_Clicking_Verify_email_button_for_unprivileged_user_results_in_permission_denied

ioigoume:
Allow Verification of Email Address, linked to CO Person, for unprivileged users.

### DIFF
--- a/app/Controller/CoInvitesController.php
+++ b/app/Controller/CoInvitesController.php
@@ -171,6 +171,23 @@ class CoInvitesController extends AppController {
                                                       filter_var($this->request->params['named']['copersonid'],FILTER_SANITIZE_SPECIAL_CHARS))));
       }
     }
+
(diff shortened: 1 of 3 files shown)


boshrin (changes requested):
A minor change: For consistency with `$managed` above and how other `isAuthorized` functions do this, `$self` should be calculated before any permissions are calculated (ie: in the same area where `$managed` is calculated) in case we decide to use it for another permission later.

ioigoume:
> A minor change: For consistency with `$managed` above and how other `isAuthorized` functions do this, `$self` should be calculated before any permissions are calculated (ie: in the same area where `$managed` is calculated) in case we decide to use it for another permission later.

Done

spartanicus (approved):
The changes have been made - and this is working well.

PR #254, opened 2021-08-30
CO-2195_Trust_CoPerson_Role_status_during_Enrollment

ioigoume:
Skip CoPersonRole::beforeSave Status calculations during an Enrollment Flow

### DIFF
--- a/app/Model/CoPersonRole.php
+++ b/app/Model/CoPersonRole.php
@@ -381,10 +381,11 @@ class CoPersonRole extends AppModel {
         $this->data[$this->alias]['valid_through'] = strftime("%F %T", $offsetDT->getTimestamp());
       }
     }
-    
+
     // If the validity of the role was changed, change the status appropriately
-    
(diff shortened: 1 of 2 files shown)


app/Model/CoPersonRole.php
```diff
@@ -383,8 +383,9 @@ public function beforeSave($options = array()) {
     }
     
     // If the validity of the role was changed, change the status appropriately
-    
-    if(!empty($this->data[$this->alias]['status'])) {
+    // Skip calculations as part of CO-2195
+    if(!empty($this->data[$this->alias]['status'])
+       && (isset($options['trustStatus']) && !$options['trustStatus'])) {
```
boshrin, line 388: I think we want to flip the logic here? `trustStatus` won't be set in most circumstances.

```suggestion
       && (!isset($options['trustStatus']) || $options['trustStatus'])) {
```

ioigoume: Done

PR #269, opened 2021-11-01
Permissions index action privileged API user group memberships (CO-2262)

skoranda:


boshrin (changes requested):
Ordinarily, it shouldn't be necessary to reference `$roles['apiuser']` in `isAuthorized()`, since the permissions are calculated in `RoleComponent::calculateCMRoles()`. So typically a privileged API User will also have `$roles['coadmin']` set.

In this case, `$managed` is set using `RoleComponent::isGroupManager()`, however the first thing that function does is return `false` if `$roles['copersonid']` is null, which it should be for an API user. The best approach would probably be to update `isGroupManager()` to check for privileged API Users, the same way that it (eventually) checks for CO Administrators.

However, the other controllers that call `isGroupMember()` do things completely differently from each other. So given how inconsistent this is we should probably just go for expediency. To that end, I think we can just test for `$roles['coadmin']` here (since it implies API User), and it'll just be slightly redundant for regular CO Admins (for whom `$managed` will be true).

PR #270, opened 2021-11-04
Improve rendering for autocomplete people picker items for easier dis…

spartanicus:
…ambiguation (CO-2263)


spartanicus:
This PR makes the layout of the CO Person autocomplete picker look like this:
image

app/Controller/CoPeopleController.php
```diff
@@ -554,18 +554,35 @@ public function find() {
...
+        }
+         
         // Make sure we don't already have an entry for this CO Person ID
         if(!Hash::check($matches, '{n}[value='.$p['CoPerson']['id'].']')) {
           $matches[] = array(
             'value' => $p['CoPerson']['id'],
-            'label' => $label
+            'label' => $label,
+            'email' => $email,
+            'emailLabel' => _txt('fd.email_address.mail'),
```
spartanicus, line 583: I'm thinking this is wrong for right-to-left languages; we should instead format the whole string via PHP text replacements and then pass that. It's both simpler and avoids i18n issues. Will convert to draft and fix this.

spartanicus:
Converted this to draft: we want to a configuration in CO Settings for what email address is considered primary for the people picker - so that only one is shown when that selection is made.

PR #279, opened 2021-11-13
CO-1336_Normalization_vs_Pipelines

ioigoume:
Do not normalize(mixCase) attributes synced to CoPersonRole from OIS


boshrin (changes requested):
This change appears to define an enumeration in the DefaultNormalizer plugin, and then references that enumeration in the core code (CoPipeline). In general, the core code should not know anything about plugin specifics. This change also breaks the interface for normalizer plugins, which is not something we can do in a minor release.

We should probably discuss this one before proceeding...

PR #280, opened 2021-11-14
CO-2283_Lightbox_View_missing_from_OIS_Name_entry_in_canvas

ioigoume:
Use lightbox to render readonly Name entry linked to OIS in canvas

### DIFF
--- a/app/Controller/NamesController.php
+++ b/app/Controller/NamesController.php
@@ -50,6 +50,16 @@ class NamesController extends MVPAController {
                            'PrimaryName'),
     'SourceName'
   );
+
+  // Use the lightbox layout for view
+  public function view($id) {
+    parent::view($id);
+    $req = $this->modelClass;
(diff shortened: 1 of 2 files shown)


app/View/Names/fields.inc
```diff
@@ -171,7 +171,7 @@
             print $this->Form->error('Name.type');
           }
         } else {
-          print $vv_available_types[ $names[0]['Name']['type'] ];
+          print _txt('en.name.type', null, $names[0]['Name']['type']);
```
boshrin, line 174: What's the reason for this change? `$vv_available_types` is the "newer" way to pass localization strings.

ioigoume: Reverted the change

boshrin:
I'm going to merge this change as is. Let's take the `$vv_available_types` discussion to a separate issue.

PR #281, opened 2021-11-14
CO-2280_View_Only_MVPAs_on_Canvas_Renders_Unformatted_in_New_Tab

ioigoume:
Format lightbox content when opened in new tab


boshrin (changes requested):
Can we use a slightly longer query parameter name than `vw`? It's going to be hard to grep for that in the code later.

PR #284, opened 2021-11-18
CO-2282_Undefined_dictionary_linked_to_attribute_Enumeration

ioigoume:
- Add hasMany AttributeEnumerations relationship to Dictionary Model
- Existence of an AttributeEnumerator is a deleteDependency

### DIFF
--- a/app/Controller/DictionariesController.php
+++ b/app/Controller/DictionariesController.php
@@ -29,7 +29,7 @@ App::uses("StandardController", "Controller");
 
 class DictionariesController extends StandardController {
   public $requires_co = true;
-  
+
   // Class name, used by Cake
   public $name = "Dictionaries";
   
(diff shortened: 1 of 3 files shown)


boshrin:
I don’t think we should delete dependent here. If an administrator tries to delete a Dictionary that is in use (eg: by an Attribute Enumeration, but also anything else that uses Dictionaries), they should get an error and then decide if they really want to delete in (in which case they have to go clean up the other entity first. Delete dependent makes it too easy to screw something up here.

ioigoume:
@boshrin the existence of an Attribute Enumerator is now a `checkDeleteDependencies` for the Dictionary Controler.

app/Controller/DictionariesController.php
```diff
@@ -40,6 +40,39 @@ class DictionariesController extends StandardController {
...
+   * @since  COmanage Registry v4.0.1
+   * @param  Array Current data
+   * @return boolean true if dependency checks succeed, false otherwise.
+   */
+
+  function checkDeleteDependencies($curdata) {
+    // We can't delete a dictionary if it is linked to an attribute enumerator
+
+    if(!empty($curdata["AttributeEnumeration"])) {
+      $deleted_status = Hash::extract($curdata, 'AttributeEnumeration.{n}.deleted');
```
boshrin, line 57: As a performance best practice, we should probably be using the various `_contains` settings on the controller. In this case, you could set `$delete_contains` to `array('AttributeEnumeration')`, and I'm not sure if `$view_contains` and `$edit_contains` need anything or can just be set to empty arrays.

app/Controller/DictionariesController.php
```diff
@@ -40,6 +40,39 @@ class DictionariesController extends StandardController {
...
+      // There are no attribute enumerators here. Proceed to delete
+      if(!in_array(false, $deleted_status)) {
+        return true;
+      }
+
+      // Return the error message
+      if($this->request->is('restful')) {
+        $this->Api->restResultHeader(403, _txt('er.dict.with.attr'));
+      } else {
+        $this->Flash->set(_txt('er.dict.with.attr'), array('key' => 'error'));
```
boshrin, line 68: Since you have the context already anyway, it might be nice to include the ID or description of the Attribute Enumeration in the error string.

`'Dictionary is in use by Attribute Enumeration X and can not be deleted'`

ioigoume: I can link a dictionary to multiple Attribute Enumerations. For the case of COUs, that can have multiple children, we return a generic message. Isn't this a similar use case?

app/Controller/DictionariesController.php
ioigoume: Done

app/Controller/DictionariesController.php
ioigoume: Done

app/Lib/lang.php
```diff
@@ -1012,6 +1012,7 @@
   'er.db.save' =>     'Database save failed',
   'er.db.save-a' =>   'Database save failed: %1$s',
   'er.dict.entry.mode' => 'Unsupported Dictionary Mode for Dictionary Entries',
+  'er.dict.with.attr' => 'Dictionary is in use by #%1$s Attribute Enumerators and can not be deleted',
```
boshrin, line 1015: Do we need the `#`?

PR #285, opened 2021-11-18
CO-2058_REST_API_Returns_500_Rather_Than_400_On_Field_Too_Wide

ioigoume:
Dynamically find and check field width

### DIFF
--- a/app/Lib/lang.php
+++ b/app/Lib/lang.php
@@ -1078,6 +1078,7 @@ original notification at
   'er.input.blank' => 'Value cannot consist of only blank characters',
   'er.input.invalid' => 'Invalid character found',
   'er.input.range' => 'Value not in permitted range',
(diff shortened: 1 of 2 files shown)


app/Lib/lang.php
```diff
@@ -1078,6 +1078,7 @@
   'er.input.blank' => 'Value cannot consist of only blank characters',
   'er.input.invalid' => 'Invalid character found',
   'er.input.range' => 'Value not in permitted range',
+  'er.input.range.len' => 'Value not in permitted range, max length permitted "%1$s"',
```
boshrin, line 1081: I would probably just call this `er.input.len`, since `range` implies something like "value must be between 1 and 10".

ioigoume: Done

PR #293, opened 2021-12-12
CO-2289_Core_API_List_of_CO_People

ioigoume:
Query CO People by their identifier.Full and only identifier response.

### DIFF
--- a/app/Lib/enum.php
+++ b/app/Lib/enum.php
@@ -785,6 +785,18 @@ class TAndCLoginModeEnum
   const DisableAllServices = 'D';
 }
 
+class TapStatusEnum
+{
+  const Accepted          = 'AC';
+  const Applied           = 'AP';
+  const Active            = 'A';
+  const Offered           = 'OF';
+  const OnLeave           = 'OL';
+  const Registered        = 'R';
+  const Suspended         = 'S';
+  const Terminated        = 'T';
+}
+
 class TemplateableStatusEnum
 {
   const Active              = 'A';
--- a/app/Lib/enum.php
+++ b/app/Lib/enum.php
@@ -785,18 +785,6 @@ class TAndCLoginModeEnum
   const DisableAllServices = 'D';
 }
 
-class TapStatusEnum
-{
-  const Accepted          = 'AC';
-  const Applied           = 'AP';
-  const Active            = 'A';
-  const Offered           = 'OF';
-  const OnLeave           = 'OL';
-  const Registered        = 'R';
-  const Suspended         = 'S';
-  const Terminated        = 'T';
-}
-
 class TemplateableStatusEnum
 {
   const Active              = 'A';
--- a/app/Plugin/CoreApi/Config/routes.php
+++ b/app/Plugin/CoreApi/Config/routes.php
@@ -25,14 +25,6 @@
  * @license       Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
  */
 
-// Define the route elements of the index action
-$core_api_read_index_route_elements =   array(
-  'plugin'     => 'core_api',
-  'controller' => 'Api',
-  'action'     => 'index',
-  '[method]'   => 'GET',
-);
-
 // The general format for Core API URLs should be /api/co/:coid/core/v1/namespace/...
 // Note REST API v2 has taken the form /api/v2/objects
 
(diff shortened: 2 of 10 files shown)


app/Lib/enum.php
```diff
@@ -785,6 +785,18 @@ class TAndCLoginModeEnum
   const DisableAllServices = 'D';
 }
 
+class TapStatusEnum
```
boshrin, line 788: Sorry, there might have been some confusion, but we're not actually supporting the TAP statuses yet. (If/when we do, we'll need to add them to CoPerson and CoPersonRole.) Strictly speaking, the Core CO Person API is _not_ a TAP API, even though it looks very similar.

app/Plugin/CoreApi/Config/routes.php
```diff
@@ -29,6 +29,7 @@
 // Note REST API v2 has taken the form /api/v2/objects
 
 // COmanage CO Person Read API
+// /api/co/:coid/core/v1/people#show
```
boshrin, line 32: Did you mean to commit these comments? (I'm not sure what the anchor tags like `#show` are referring to.)

app/Plugin/CoreApi/Config/routes.php
```diff
@@ -50,7 +64,13 @@
   )
 );
 
+// XXX Scoped identifiers are not parsed properly because they are perceived as file extensions
```
boshrin, line 67: Ah yes, definitely a problem. But why is this still flagged `XXX`?

app/Plugin/CoreApi/Controller/ApiController.php
```diff
@@ -186,31 +201,150 @@ public function beforeFilter() {
...
+    try {
+      $query_filters = array();
+      // Load the default ordering and pagination settings
+      $this->Paginator->settings = $this->paginate;
+      $this->Paginator->settings['conditions']['Identifier.type'] = $this->cur_api['CoreApi']['identifier_type'];
+      $this->Paginator->settings['conditions']['Identifier.deleted'] = false;
+      $this->Paginator->settings['conditions']['Identifier.identifier_id'] = null;
+      $this->Paginator->settings['conditions']['Identifier.status'] = SuspendableStatusEnum::Active;
+      $this->Paginator->settings['conditions']['CoPerson.co_id'] = $this->cur_api['CoreApi']['co_id'];
+      // We allow people of any status to be pulled, though maybe we could offer a filter
```
boshrin, line 230: This comment isn't needed anymore?

app/Plugin/CoreApi/Lib/lang.php
```diff
@@ -41,17 +41,25 @@
     CoreApiEnum::CoPersonRead     => 'COmanage CO Person Read API',
     CoreApiEnum::CoPersonWrite    => 'COmanage CO Person Write API'
   ),
-  
+
+  'pl.coreapi.en.response.type' => array(
+    ResponseTypeEnum::Full              => 'Full Profile(s)',
```
boshrin, line 46: I'd wordsmith these to `Full Profile` and `Identifier Only`.

app/Plugin/CoreApi/Lib/lang.php
```diff
@@ -41,17 +41,25 @@
...
-  'pl.coreapi.identifier.desc' => 'The Identifier type used to map API locater identifiers to CO Person records',
-  'pl.coreapi.info'            => 'The Core API endpoint is %1$s',
-  'pl.coreapi.rs.edited-a4'    => '%1$s Edited via Core API: %2$s',
-  'pl.coreapi.rs.linked'       => 'New Org Identity Linked via Core API',
+  'pl.coreapi.api'                  => 'API',
+  'pl.coreapi.api_user.desc'        => 'The API User authorized to make requests to this endpoint',
+  'pl.coreapi.identifier.desc'      => 'The Identifier type used to map API locate identifiers to CO Person records',
+  'pl.coreapi.info'                 => 'The Core API endpoint is %1$s',
+  'pl.coreapi.response.type'        => 'Response Type',
+  'pl.coreapi.response.type.desc'   => 'Define the response data granularity',
```
boshrin, line 62: I'd say `response content`.

app/Plugin/CoreApi/Model/CoreApi.php
```diff
@@ -26,18 +26,36 @@
...
   // Add behaviors
   public $actsAs = array('Changelog', 'Containable');
-  
+
+  // List of allowed query parameters
+  // 1. range, inList, comparison are all validation rules supported by CAKE core
+  //    In runtime it will be translated to Validation::validation_rule, Validation::email
+  // 2. CAKEPHP 2.x uses lowercased and underscored URLS (InflectedRoute class in CAKEPHP 3.0)
+  //    as a result person.status becomes person_status
+  const ALLOWED_QUERY_PARAMS = array(
```
boshrin, line 47: Is there a reason to define this as `const` rather than a private/protected var?

app/Plugin/CoreApi/Model/CoreApi.php
```diff
@@ -113,8 +131,36 @@ class CoreApi extends AppModel {
...
+   * @since  COmanage Registry v4.1.0
+   */
+
+  public function beforeValidate($options = array())
+  {
+    if(empty($this->data['CoreApi']["api"])) {
+      return false;
+    }
+    // Disable Response type for Write/Delete actions
+    if($this->data['CoreApi']["api"] === CoreApiEnum::CoPersonWrite) {
```
boshrin, line 156: The Write API implies access to the Read API, so I think we still want the `index_response_type` setting.

boshrin (changes requested):

app/Plugin/CoreApi/Config/routes.php
ioigoume: No. I will remove them.

app/Plugin/CoreApi/Config/routes.php
ioigoume: Good point. I will remove this.

app/Plugin/CoreApi/Controller/ApiController.php
ioigoume: done

app/Plugin/CoreApi/Lib/lang.php
ioigoume: Done

app/Plugin/CoreApi/Lib/lang.php
ioigoume: Done

app/Plugin/CoreApi/Model/CoreApi.php
ioigoume: Thank you for the clarification on that!

app/Plugin/CoreApi/Model/CoreApi.php
ioigoume: No. Will change.

app/Plugin/CoreApi/Model/CoreApi.php
ioigoume: Done

app/Lib/enum.php
ioigoume: Done

app/Plugin/CoreApi/Model/CoreApi.php
```diff
@@ -26,18 +26,33 @@
...
   public $actsAs = array('Changelog', 'Containable');
-  
+
+  // List of allowed query parameters
+  // 1. range, inList, comparison are all validation rules supported by CAKE core
+  //    In runtime it will be translated to Validation::validation_rule, Validation::email
+  // 2. https://www.php.net/manual/en/language.variables.external.php#language.variables.external.dot-in-names
+  //    Currently the dot character is not a valid character for a PHP variable name. So co_person.status
+  //    will be transformed to co_person_status.
+  // XXX We are making the convention that the field will be placed last and will be a single word
```
boshrin, line 48: I feel like this will be a problem eventually, but it seems our options are limited.

PR #296, opened 2021-12-16
CO-1904_First_field_does_not_have_input_focus_when_manually_adding_new_CO_person_role

ioigoume:
Enable autofocus for the first field of CoPersonRole add/view form

### DIFF
--- a/app/View/CoPersonRoles/fields.inc
+++ b/app/View/CoPersonRoles/fields.inc
@@ -156,7 +156,7 @@
         'CoPersonRole.title'
       )
     );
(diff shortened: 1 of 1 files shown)


spartanicus:
Thanks Ioannis - note that if there are no COUs, we need to focus the Affiliation instead - can we update this to include that?

ioigoume:
@spartanicus suggestion is now implemented

spartanicus:
The move to the HTML5 attribute is a particularly good call (the original implementation of this goes back a while). Merging.

PR #298, opened 2021-12-28
CO-2314_Core_Write_API_Delete

ioigoume:


app/Config/Schema/schema.xml
```diff
@@ -2836,6 +2836,9 @@
     <field name="actor_co_person_id" type="I">
       <constraint>REFERENCES cm_co_people(id)</constraint>
     </field>
+    <field name="actor_api_user_id" type="I">
+      <constraint>REFERENCES cm_api_users(id)</constraint>
+    </field>
```
boshrin, line 2841: This is something we probably should have had all along. The challenge is that to do this properly we'd have to go back and update every place where a history record might be created by an API user, which might be a lot of work. Or maybe it's mostly StandardController.

PR #300, opened 2022-01-12
Update RoleComponent.php

ioigoume:
Handle null values of `deleted` column.


skoranda:
Tested against deployment using MySQL.

PR #305, opened 2022-01-21
[NO JIRA] Error message key is not present in lang.php

ioigoume:
`er.fields.api.frozen` is used by AppModel.php but is missing from lang.php


boshrin:
I removed that `beforeValidate` code in e140fadd37 as part of CO-2294. It looks like it got dragged back in as part of a946724733 (PR #282, CO-2294). Can you review to see if anything else was unintentionally reverted?

ioigoume:
Closed in favor of [Pull Request 309](https://github.com/Internet2/comanage-registry/pull/309)

PR #308, opened 2022-01-24
CO-2336_Rest_API_v1_index_requests_for_CO_user_fail

ioigoume:
Parse CO ID from query params


boshrin:
I just tested `hotfix-4.0.x` _without_ this patch and a query like `GET /registry/cous.json?coid=138` works (authenticating as a CO specific API user). What URL do you have that breaks?

PR #312, opened 2022-02-04
(cleaner version submitted, ignore)

allankim:
Handle middle name in ApiSource


boshrin:
Hey @allankim, we can accept this, but can you resubmit it without all the merge noise?

allankim:
Submitting cleaner version

PR #315, opened 2022-02-08
CO-2346_privileged_API_user_cannot_edit_CoGroup_in_CO

ioigoume:
Fix api user CoGroup priviledges


boshrin:
It shouldn't be necessary to check for `apiuser` in `isAuthorized`. `RoleComponent::calculateCMRoles()` will set `cmadmin` or `coadmin` for API users as needed. I think what this patch is doing is effectively bypassing the `$managed` check for API Users.

The proper solution is probably for `isGroupMember` to know how to handle API Users, but that will require a significant rewrite that isn't worth the effort for v4. So maybe the compromise is to calculate `$managed` based on something like `($roles['apiuser'] && $roles['coadmin']) || isGroupManager()`.

app/Controller/Component/RoleComponent.php
```diff
@@ -1346,7 +1346,20 @@ public function isCouPerson($coPersonId, $coId, $couId) {
...
-    if(!$coPersonId || !$coGroupId) {
+    if(!$coGroupId) {
+      return false;
+    }
+
+    if($coPersonId === false) {
+      // The API User can be CO admin only if the co_id is equal to the
+      // co_id we calculated from the request data. In order for this
+      // to happen the Group and the API User must belong to the same CO
+      $roles = $this->calculateCMRoles();
```
boshrin, line 1357: `calculateCMRoles()` isn't really intended to be called recursively, it's more for Controllers to call. I'd be a little concerned about side effects since this is old code. It might make more sense to just directly implement the logic you need here, eg:

```
if($this->Session->check('Auth.User.api')) {
  if($authCoId == 1 || (($coId == $authCoId) && $privd)) {...}
}
```

Or maybe refactor that particular section from `calculateCMRoles` into its own function that both `calculateCMRoles` and `isGroupManager` can call.

boshrin:
I believe this wins for simplest, good enoughest approach.

PR #317, opened 2022-02-11
CO-2353_REST_users_cannot_create_UnixCluster

ioigoume:
Fix UnixCluster version and add permissions for priviledged REST API users


skoranda:
The commit has been changed to only add the version so I am going to merge.

PR #334, opened 2022-03-10
Correct CoPersonRole for approval notifications (CO-2382)

skoranda:


boshrin:
Is this for an additional role enrollment? If so, are there other cases where `$pt['EnrolleeCoPerson']['CoPersonRole'][0]` would potentially refer to the wrong CoPersonRole?

skoranda:
> Is this for an additional role enrollment? 

NEW_COU is most commonly needed during an additional role enrollment, yes.

> If so, are there other cases where `$pt['EnrolleeCoPerson']['CoPersonRole'][0]` would potentially refer to the wrong CoPersonRole?

I don't understand the question. Are you asking if there are other places in the code that should be investigated to see if `$pt['EnrolleeCoPerson']['CoPersonRole'][0]` is not correct?

boshrin:
> I don't understand the question. Are you asking if there are other places in the code that should be investigated to see if `$pt['EnrolleeCoPerson']['CoPersonRole'][0]` is not correct?

Specifically in the enrollment context (`Model/CoPetition.php` or related code).

boshrin:
I manually merged this as `d14bfcd1be` on `develop` in order to fold in another change.

PR #338, opened 2022-03-20
CO-2371_Identifier_filter_on_My_Population_page_returns_duplicate_results

ioigoume:
Remove MyPopulation search duplicates when a search for partial attribute value fetches the same result more than once


spartanicus:
I see no issue with this from a UI perspective, and the pagination behaves as it should.

boshrin:
I think this is good enough for now, but I'd probably aim for a more generalized solution in PE.

PR #341, opened 2022-03-21
CO-2323_CoreAPI_create

ioigoume:
CoreAPI Create User

### DIFF
--- a/app/Model/AppModel.php
+++ b/app/Model/AppModel.php
@@ -1206,7 +1206,76 @@ class AppModel extends Model {
     
     return false;
   }
-  
+
+  /**
+   * This is copied by the framework itself but we take into consideration the soft delete functionality COmanage
+   * has in place.
+   *
+   * Returns false if any fields passed match any (by default, all if $or = false) of their matching values.
+   *
+   * Can be used as a validation method. When used as a validation method, the `$or` parameter
+   * contains an array of fields to be validated.
+   *
+   * @param array $fields Field/value pairs to search (if no values specified, they are pulled from $this->data)
+   * @param bool|array $or If false, all fields specified must match in order for a false return value
+   * @return bool False if any records matching any fields are found
+   */
+  public function isUniqueChangelog($fields, $or = true) {
+    if (is_array($or)) {
+      $isRule = (
+        array_key_exists('rule', $or) &&
+        array_key_exists('required', $or) &&
+        array_key_exists('message', $or)
+      );
+      if (!$isRule) {
+        $args = func_get_args();
+        $fields = $args[1];
+        $or = isset($args[2]) ? $args[2] : true;
+      }
+    }
+    if (!is_array($fields)) {
+      $fields = func_get_args();
+      $fieldCount = count($fields) - 1;
+      if (is_bool($fields[$fieldCount])) {
+        $or = $fields[$fieldCount];
+        unset($fields[$fieldCount]);
+      }
+    }
+
+    foreach ($fields as $field => $value) {
+      if (is_numeric($field)) {
+        unset($fields[$field]);
+
+        $field = $value;
+        $value = null;
+        if (isset($this->data[$this->alias][$field])) {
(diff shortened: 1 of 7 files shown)


app/Model/AppModel.php
```diff
@@ -1195,7 +1195,76 @@ public function isPlugin($type = null) {
...
+   * Returns false if any fields passed match any (by default, all if $or = false) of their matching values.
+   *
+   * Can be used as a validation method. When used as a validation method, the `$or` parameter
+   * contains an array of fields to be validated.
+   *
+   * @param array $fields Field/value pairs to search (if no values specified, they are pulled from $this->data)
+   * @param bool|array $or If false, all fields specified must match in order for a false return value
+   * @return bool False if any records matching any fields are found
+   */
+  public function isUniqueChangelog($fields, $or = true) {
```
boshrin, line 1212: I'm not sure why we're adding this check? We support multiple copies of the same EmailAddress or Identifier on the same record, for example this is a common use case when syncing records from multiple Org Identity Sources.

app/Plugin/CoreApi/Config/routes.php
```diff
@@ -71,16 +71,10 @@
...
-// Scoped identifiers are not parsed properly because they are perceived as file extensions
-// Enable extensions parse to resolve this problem
-Router::parseExtensions('json', 'xml');
 
-
-/*
-
-Router::connect(
 // XXX This needs to trigger identifier assignment and maybe some other stuff
 //     provisioning should only fire after all models are saved
```
boshrin, line 76: It looks like you did this, so we can probably delete this comment?

app/Plugin/CoreApi/Model/CoreApi.php
```diff
@@ -180,6 +180,214 @@ public function cmPluginMenus() {
...
+      // Assign Identifiers
+      // Disable provisioning since we will bulk provision at the end
+      // This will return an array describing which, if any, identifiers were assigned,
+      // but we don't do anything with the result here
+      $this->Co->CoPerson->Identifier->assign('CoPerson', $co_person_id, null, false, $actorApiUserId);
+
+      // Handle plugin models
+      // - Authenticator
+      // - XXX Clusters currently not supported
+      $not_supported_plugins = array('UnixCluster');
```
boshrin, line 342: Why aren't clusters supported? And why is only the UnixCluster plugin excluded?

app/Plugin/CoreApi/Controller/ApiController.php
```diff
@@ -211,6 +212,31 @@ public function beforeFilter() {
...
 
+  /**
+   * Handle a Core API CO Person record create
+   *
+   * @since  COmanage Registry v4.1.0
+   */
+
+  public function create() {
+    try {
+      $ret = $this->CoreApi->createV1($this->cur_api['CoreApi']['co_id'],
```
boshrin, line 223: `createV1` doesn't appear to actually return anything? We need to convey the Identifier of the configured Identifier Type in the response.

app/Plugin/CoreApi/Config/routes.php
ioigoume: Done

app/Plugin/CoreApi/Controller/ApiController.php
ioigoume: done

app/Plugin/CoreApi/Model/CoreApi.php
ioigoume: I am not sure why i added this check there. I removed  it.

app/Model/AppModel.php
ioigoume: @boshrin let's discuss this over the Dev call.

ioigoume: @boshrin Removed the `isUniqueChangelog` validation rule from the `EmailAddress` and the `Url` Model

PR #356, opened 2022-04-06
CO-2416_ActorApiUser_is_missing_from_Identifier_Assignment_via_Namespace_API_impelementation

ioigoume:
Add Actor Api User support and missing $db variable initialization in IdentifierAssignment Model

### DIFF
--- a/app/Model/CoIdentifierAssignment.php
+++ b/app/Model/CoIdentifierAssignment.php
@@ -187,7 +187,7 @@ class CoIdentifierAssignment extends AppModel {
    * @throws RuntimeException
    */
   
-  public function assign($coIdentifierAssignment, $objType, $objId, $actorCoPersonID, $provision=true, $actorApiUserId=null) {
+  public function assign($coIdentifierAssignment, $objType, $objId, $actorCoPersonId, $provision=true, $actorApiUserId=null) {
     $ret = null;
 
     // Determine if we are actually assigning an email address instead of an identifier.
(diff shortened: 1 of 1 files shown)


app/Model/CoIdentifierAssignment.php
```diff
@@ -255,6 +255,7 @@ public function assign($coIdentifierAssignment, $objType, $objId, $actorCoPerson
     
     if($coIdentifierAssignment['CoIdentifierAssignment']['algorithm'] == IdentifierAssignmentEnum::Plugin) {
       if(empty($coIdentifierAssignment['CoIdentifierAssignment']['plugin'])) {
+        $dbc->commit();
```
boshrin, line 258: I think these `commit`s should be `rollback`s?

app/Model/CoIdentifierAssignment.php
```diff
@@ -499,7 +502,14 @@ private function checkInsert($coIdentifierAssignment, $objType, $assignEmail, $o
     
     $coId = $coIdentifierAssignment['CoIdentifierAssignment']['co_id'];
     $fk = Inflector::underscore($objType) . "_id";
-    
+
+
+    // Begin a transaction. This is more because we need to ensure the integrity of
+    // data between SELECT and INSERT/UPDATE than that we expect to rollback.
+
+    $dbc = $this->getDataSource();
```
boshrin, line 510: Why are we starting a new transaction here? `checkInsert()` should be called from within a transaction.

ioigoume: @boshrin thank you for the clarification. `checkInsert()` is using `$dbc->rollback()` but there is no initialization of the `$dbc`. I wrongly assumed that the function should handle the transaction by itself. Following your comment, i removed the transaction initialization and passed the `$dbc` variable as a parameter. Now `checkInsert()` has a reference to the transaction and can call a rollback if an exception occurs.

boshrin: `checkInsert()` doesn't need to call the rollback. If it throws an Exception, the parent code will catch it and call `rollback()`.

ioigoume: Done

PR #357, opened 2022-04-09
CO-2417_CoGroup_Select_action_should_enforce_copersonid_named_parameter_presence

ioigoume:
Default to current user if copersonid named param is missing for CoGroups Select action


app/Controller/CoGroupsController.php
```diff
@@ -753,7 +753,9 @@ public function paginationConditions() {
     // First get the CoPersonID to work against either the current user, or the user being acted on in select mode
     $coPersonId = $this->Session->read('Auth.User.co_person_id');
     if($this->action == 'select') {
```
spartanicus, line 755: For simplicity, might we just add the test for a non-empty coPersonId here?

ioigoume: Done

PR #359, opened 2022-04-22
Style CO Petitions Duplicate Check (CO-1796)

spartanicus:
This PR formats and styles the CO Petition Duplicate Check against a Match server to look like the following:
image


app/Lib/lang.php
```diff
@@ -134,6 +134,7 @@
   'ct.co_people.pl' =>          'CO People',
   'ct.co_person_roles.1' =>     'CO Person Role',
   'ct.co_person_roles.pl' =>    'CO Person Roles',
+  'ct.co_petition.duplicate.check' => 'CO Petition Duplicate Check',
```
boshrin, line 137: This isn't a controller localization, this key should be elsewhere, probably `op.`.

app/Lib/lang.php
```diff
@@ -1672,6 +1677,7 @@
   'fd.name.primary_name' => 'Primary',
   'fd.name.fields.req' => 'A name must consist of at least these fields:',
   'fd.nested' => 'Nested',
+  'fd.new' =>           'New',
```
boshrin, line 1680: `fd.match.new`?

app/Lib/lang.php
```diff
@@ -2064,6 +2070,7 @@
   'op.add' =>         'Add',
   'op.add-a' =>       'Add %1$s',
   'op.add.new' =>     'Add a New %1$s',
+  'op.add.new.literal' => 'Add new',
```
boshrin, line 2073: `op.match.new`?

app/Lib/lang.php
```diff
@@ -2186,7 +2193,10 @@
   'op.manage' =>      'Manage',
   'op.manage-a' =>    'Manage %1$s',
   'op.manage.grm' =>  'Manage Group Memberships',
+  'op.match.add.confirm' => 'Are you sure you wish to add a new CO Person record?',
```
boshrin, line 2196: "Are you sure you want to create a new CO Person record"?

app/Lib/lang.php
```diff
@@ -2186,7 +2193,10 @@
...
+  'op.match.merge' => 'Merge',
   'op.match.new' =>   'Assign a new Reference Identifier',
+  'op.match.select.confirm' => 'Do you want to merge the new identity with this existing identity?',
```
boshrin, line 2199: "Are you sure you want too..."?

boshrin (changes requested):
Proposed some wordsmithing.

spartanicus:
Thanks Benn - the text changes have been put in place.

app/Lib/lang.php
spartanicus: The lang keys have been reworked into op.match and fd.match. 
This one is "op.match.duplicate.check"

app/Lib/lang.php
spartanicus: Changed as suggested.

app/Lib/lang.php
spartanicus: Changed as suggested.

app/Lib/lang.php
spartanicus: Changed as suggested.

PR #360, opened 2022-04-22
Make user and realm available to TOTP  Authenticator for labeling     (CO-2418)

shaynasings:
(CO-2418)

User and Realm do not display on TOTP label in Google Authenticator even though specified by PrivacyIdea policy.


app/AvailablePlugin/PrivacyIdeaAuthenticator/Model/PrivacyIdea.php
```diff
@@ -171,9 +171,11 @@ public function createToken($privacyIdeaAuthenticator, $coPersonId) {
     $params = array(
       'type' => 'totp',
       'genkey' => '1',
+      'user' => $identifier,
+      'realm' => $privacyIdeaAuthenticator['realm'],
       'optlen' => '6'
-    );
-    
+    );    
+
```
boshrin, line 178: Since this whitespace change isn't related to the fix, it shouldn't be included in the PR.

PR #363, opened 2022-04-29
defaulting to blank dropdown option

rweber87:
Ticket: https://todos.internet2.edu/browse/CO-2392

Comments:
defaulting to blank option if none is specified when page loads
165752087-24bee0d3-dbff-46a0-b6ed-0fa14ccc5d4c

### DIFF
--- a/app/Controller/CoPersonRolesController.php
+++ b/app/Controller/CoPersonRolesController.php
@@ -319,7 +319,7 @@ class CoPersonRolesController extends StandardController {
         
         $cous = $this->CoPersonRole->Cou->find('threaded', $args);
         
-        $childCous = array();
+        $childCous = array('' => '');
         foreach($cous as $cou) {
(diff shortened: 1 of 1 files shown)


spartanicus:
Hi, Robbie! So - it isn't necessary to set the array values to a blank string for initialization -- the 'empty' field attribute set a few lines below takes care of that: `$attrs['empty'] = $vv_allow_empty_cou;` . i.e. if 'empty' is true - based on the boolean $vv_allow_empty_cou - the first select option will automatically be blank (or a string placeholder if set ... in this case it's just blank). 

However, when the $vv_allow_empty_cou is true, the field needn't be marked as required with the red asterisk (*). Can you remove the blank string initialization and change the code so that the required asterisk doesn't show when empty cous are allowed?

shaynasings:
closing per @spartanicus

PR #368, opened 2022-05-10
Fix default pattern selection in CoIdentifierAssignements (CO-2391)

spartanicus:
This PR sets the default format value to "(#)", ensures that selecting the default in the drop-down sets the field, and removes the empty option. 

image


spartanicus:
Q: should "description" or "format" be required fields?

PR #382, opened 2022-06-15
Generate correct User-Menu plugin links and allow plugin path to be overridden (CO-2312)

spartanicus:
This PR should allow plugins to better assert links into the user menu.

### DIFF
--- a/app/View/Elements/menuUser.ctp
+++ b/app/View/Elements/menuUser.ctp
@@ -217,6 +217,8 @@
 
         // Plugin submenus
         // This rendering is a bit different from how render_plugin_menus() does it...
+        // Each plugin will appear by name (e.g. "Announcements") with all appropriate 
+        // COs listed beneath it. Each CO title will link to the plugin within that CO.
         if(!empty($menuContent['plugins'])) {
           $userPluginsExist = false;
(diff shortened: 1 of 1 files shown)


app/View/Elements/menuUser.ctp
```diff
@@ -241,11 +243,19 @@
...
                       }
   
+                      // Get the plugin link array
                       $args = $menuContent['plugins'][$plugin]['coperson'][$label];
   
+                      // Always include the co_person_id and the co_id
                       $args[] = 'copersonid:' . $co['co_person_id'];
-                      $args['plugin'] = Inflector::underscore($plugin);
-  
+                      $args[] = 'co:' . $co['co_id'];
```
boshrin, line 251: Shouldn't this be `$args['co'] => $co['co_id']`? In which case the `copersonid` line above it is also wrong...

spartanicus: Yes - you're correct. It was wrong in the original source, and I thoughtlessly followed suit. Fixing...

spartanicus: Code style cleanup pushed (and current code rebased against the latest develop).

PR #385, opened 2022-06-19
[NO JIRA] Button_spinner_display

ioigoume:
Correctly align button working spinner

### DIFF
--- a/app/View/CoPeople/expunge.ctp
+++ b/app/View/CoPeople/expunge.ctp
@@ -192,16 +192,25 @@
         'span',
         '',
         array(
-          'class' => 'spinner-grow spinner-grow-sm mr-1 align-middle invisible btn-submit-with-loader',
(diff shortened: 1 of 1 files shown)


app/View/CoPeople/expunge.ctp
```diff
@@ -192,16 +192,26 @@
         'span',
         '',
         array(
-          'class' => 'spinner-grow spinner-grow-sm mr-1 align-middle invisible btn-submit-with-loader',
+          'class' => 'spinner-grow spinner-grow-sm mr-2 align-middle invisible btn-submit-with-loader',
           'escape' => false,
           'role' => 'status',
           'aria-hidden' => true
```
spartanicus, line 198: Set the aria-hidden attribute to 'true' (string) rather than true (= 1 in output). The attribute can only take the string value.

spartanicus (approved):

PR #394, opened 2022-07-06
CO-2458 Add support for Username Reminder and Identifier substitution for Aut…

shaynasings:
…henticator message templates (SMU PMO958)


app/AvailablePlugin/PasswordAuthenticator/Controller/PasswordsController.php
```diff
@@ -221,7 +231,7 @@ public function ssr() {
             
             // We render success but let the form render again anyway, in case the
             // user wants to try again.
-            $this->Flash->set(_txt('pl.passwordauthenticator.ssr.sent'), array('key' => 'success'));
+            $this->Flash->set(_txt('pl.passwordauthenticator.ssrorreminder.sent'), array('key' => 'success'));
```
boshrin, line 234: I don't think we need to rename the `.ssr` language keys (here or below).

app/AvailablePlugin/PasswordAuthenticator/Model/PasswordResetToken.php
```diff
@@ -205,12 +205,19 @@ public function generateRequest($authenticatorId, $q) {
       throw new InvalidArgumentException(_txt('er.notfound', array(_txt('ct.authenticators.1'), $authenticatorId)));
     }
     
-    // Next, generate a token in order to send it
-    $token = $this->generate($passwordAuthenticator['PasswordAuthenticator']['id'], $coPersonId);
+    // Get the name of the calling function to determine whether we're sending Reset Token Request or Username Reminder
+    $trace = debug_backtrace();
```
boshrin, line 209: Instead of using debug_backtrace, add a parameter to the function signature indicating how it's supposed to be called. eg: `public function generateRequest($authenticatorId, $q, $mode='reset')`

app/AvailablePlugin/PasswordAuthenticator/Model/PasswordResetToken.php
```diff
@@ -260,9 +251,60 @@ protected function send($passwordAuthenticator, $coPersonId, $token, $recipients
...
+   * @param  array                 $coPerson              CO Person with associated email addresses and identifiers
+   * @param  array                 $recipients            Array of email addresses to send token to
+   * @param  int                   $actorCoPersonId       Actor CO Person ID
+   * @param  array		   $substitutions	  substitutions for the message template
+   * @throws InvalidArgumentException
+   */
+  protected function send($passwordAuthenticator, $coPerson, $recipients, $actorCoPersonId, $substitutions=null) {
+
+    // If $substitutions are null, this is a username reminder. Set the message template type and text for history record, to be used later, appropriately based on this
+    if ($substitutions == null) {
```
boshrin, line 278: Checking for `$substitutions` being NULL is brittle. Instead, either pass a mode parameter, or have the caller provide the message template and history text.

Or since we already have two functions anyway (`send` and `sendRequest`), just have two separate functions - one for reset and the other for reminder.

app/AvailablePlugin/PasswordAuthenticator/Lib/lang.php
```diff
@@ -86,15 +86,24 @@
...
+  'pl.passwordauthenticator.ssr.mt'         => 'Reset Message Template',
   'pl.passwordauthenticator.ssr.mt.desc'    => 'Message template used for email to send reset instructions to',
-  'pl.passwordauthenticator.ssr.q'          => 'Email Address or Identifier',
   'pl.passwordauthenticator.ssr.redirect'   => 'Redirect on Self Service Reset',
   'pl.passwordauthenticator.ssr.redirect.desc' => 'URL to redirect to on successful self service reset',
-  'pl.passwordauthenticator.ssr.sent'       => 'An email with further instructions has been sent to the address on record',
   'pl.passwordauthenticator.ssr.url'        => 'Self Service Reset Initiation URL',
   'pl.passwordauthenticator.ssr.validity'   => 'Self Service Reset Token Validity',
   'pl.passwordauthenticator.ssr.validity.desc' => 'Time in minutes the reset token is valid for',
+  'pl.passwordauthenticator.ssrorreminder.q' => 'Email Address or Identifier',
```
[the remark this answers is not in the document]
shaynasings: Ok, should I add an identical key for .reminder? Or should I reuse the ssr keys (even though they would be used for remind)? I'm not aware of all the implications here.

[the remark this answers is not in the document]
boshrin: It's the same text for the same context, we don't need two keys, especially since it's in a plugin so if we do need to break the keys apart later for some reason it's not a big deal.

PR #395, opened 2022-07-09
CO-2456_Core_API_READ_fails_when_Identifier_is_a_URL

ioigoume:
urldecode the person identifier

### DIFF
--- a/app/Plugin/CoreApi/Config/routes.php
+++ b/app/Plugin/CoreApi/Config/routes.php
@@ -40,6 +40,30 @@ Router::connect(
   )
 );
 
+// COmanage CO Person API DELETE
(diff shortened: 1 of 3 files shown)


boshrin:
If I'm understanding this correctly, this approach will break existing deployments and require all identifiers to be base64 encoded. I thought we only wanted to require base64 encoding when there was a special character in the string.

We could either (1) check to see if the string is a valid identifier, and if it isn't only then base64 decode it, or (2) add some sort of special notation (which itself might need to be url encoded) to indicate the identifier needs decoding. For the second option, maybe something like `/v1/people/base64:ENCODEDSTRING`?

boshrin:
As per discussion from this week's dev meeting, I've added some notes to [CO-2456](https://todos.internet2.edu/browse/CO-2456).

PR #399, opened 2022-07-22
Fix validation of HttpServer model (CO-2467)

skoranda:


spartanicus:
Tested. This fix looks good to me.

PR #403, opened 2022-08-17
Self-Service Email Dashboard Widget Plugin (CO-2449)

spartanicus:


app/AvailablePlugin/EmailWidget/Model/EmailWidgetEmail.php
```diff
@@ -0,0 +1,166 @@
...
+   * along with the email address and email type to be  added.
+   * This is step one of a two-step email verification process.
+   * 
+   * @since COmanage Registry v4.1.0
+   * @param string $emailAddress Email address to be added
+   * @param string $emailType    Type of email address to be added 
+   * @return integer id of the new row
+   */
+  public function generateToken($emailAddress, $emailType, $primary) {
+    $token = bin2hex(random_bytes(3));
```
boshrin, line 61: This should call `generateRandomToken` (available in `Lib/util.php`). While that function was originally written because `random_bytes` wasn't added until PHP 7 (and at the time we were supporting PHP 5.4), it also skips l and 1 to avoid confusion and inserts dashes into longer tokens.

app/AvailablePlugin/EmailWidget/Model/EmailWidgetEmail.php
```diff
@@ -0,0 +1,166 @@
...
+   * @param string $token Token used for verification
+   * @param string $id row id for token lookup
+   * @return string outcome of verification
+   */
+  public function verify($token, $id, $coPersonId) {
+    $args = array();
+    $args['conditions']['id'] = $id;
+    $rec = $this->find('first',$args);
+    if($rec['EmailWidgetEmail']['token'] == $token) {
+      // If more than 10 minutes have elapsed (600 seconds) fail with "timeout".
```
boshrin, line 97: For comparison, CoInvites (which will be heavily refactored) stores the expiration time in the invitation.

app/Lib/lang.php
```diff
@@ -1287,6 +1287,7 @@
   'er.ug.version' =>  'Unknown version "%1$s"',
   'er.unknown' =>     'Unknown value "%1$s"',
   'er.unknown.def' =>  'Unknown value',
+  'er.unknown.er' =>  'An unknown error occurred.',
```
boshrin, line 1290: How do we have an unknown error? Don't we control everything here? (If I'm an admin, what do I do with this? If I'm a COmanage developer, how do I debug it?)

boshrin (changes requested):

app/AvailablePlugin/EmailWidget/Config/Schema/schema.xml
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0" ?>
+<!--
+  COmanage Registry Self Service Email Widget Plugin Database Schema
```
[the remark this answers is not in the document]
spartanicus: No worries - I like nitpicky. I'll get these modified.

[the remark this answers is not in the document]
spartanicus: Done. Note that this change is reflected all the way up and down the code - so if you've previously installed this plugin, you'll need to reinstall it because its name has changed even at the plugin directory level.

app/AvailablePlugin/EmailWidget/Config/Schema/schema.xml
```diff
@@ -0,0 +1,69 @@
...
+<schema version="0.3">
+  <table name="co_email_widgets">
+    <field name="id" type="I">
+      <key />
+      <autoincrement />
+    </field>
+    <field name="co_dashboard_widget_id" type="I">
+      <constraint>REFERENCES cm_co_dashboard_widgets(id)</constraint>
+    </field>
+    <field name="created" type="T" />
```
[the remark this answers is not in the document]
spartanicus: Done

app/AvailablePlugin/EmailWidget/Config/Schema/schema.xml
```diff
@@ -0,0 +1,69 @@
...
+    <field name="modified" type="T" />
+    <field name="email" type="C" size="128" />
+    <field name="type" type="C" size="32" />
+    <field name="primary_email" type="L" />
+    <field name="token" type="C" size="8" />
+
+    <index name="email_widget_emails_i1">
+      <col>id</col>
+      <unique />
+    </index>
```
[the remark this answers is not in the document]
spartanicus: The index has been removed.

app/AvailablePlugin/EmailWidget/Lib/lang.php
```diff
@@ -0,0 +1,89 @@
...
+  // Plugin texts
+  'pl.email_widget.add' => 'Add an email address',
+  'pl.email_widget.added' => 'Email address added',
+  'pl.email_widget.email.subject' => 'Your email verification code',
+  'pl.email_widget.email.body' => 'Please copy the following code into the Email Address panel in COmanage Registry to verify your email address.',
+  'pl.email_widget.error.status' => 'HTTP Status: ',
+  'pl.email_widget.error.message' => 'Message: ',
+  'pl.email_widget.error.401' => 'Your session has timed out. Please log in.',
+  'pl.email_widget.error.500' => 'There was an unknown error.',
+  'pl.email_widget.error.invalid' => 'Please enter a valid email address.',
```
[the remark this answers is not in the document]
spartanicus: Currently, I've changed this to `email_address_widget` - do you want this changed to the version without underscores (`emailaddresswidget`) or should we leave it like this?

app/AvailablePlugin/EmailWidget/Config/Schema/schema.xml
```diff
@@ -0,0 +1,69 @@
...
+      <key />
+      <autoincrement />
+    </field>
+    <field name="co_dashboard_widget_id" type="I">
+      <constraint>REFERENCES cm_co_dashboard_widgets(id)</constraint>
+    </field>
+    <field name="created" type="T" />
+    <field name="modified" type="T" />
+    <field name="default_type" type="C" size="128" />
+    <field name="hide_type" type="L" />
```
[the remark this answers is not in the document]
spartanicus: This was added when I believed we might let an administrator choose to show the type selector to end users when adding or editing an email address. That's not currently a feature - an end user can not select the type at all right now.  Given the way we're using the widget to enforce a default, perhaps it should not be a feature for self-service end users. Should we keep this for a future enhancement, or should we just remove it? 

A better name might be _type_hidden_, though, and I'll change this field to that for now.

app/AvailablePlugin/EmailWidget/Lib/lang.php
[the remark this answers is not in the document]
spartanicus: ...However I'll go ahead and change this to be consistent with the smushed-together style of language keys (`emailaddresswidget`).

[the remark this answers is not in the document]
spartanicus: Done and pushed.

app/AvailablePlugin/EmailWidget/Config/Schema/schema.xml
```diff
@@ -0,0 +1,69 @@
...
+      <autoincrement />
+    </field>
+    <field name="email_widget_email_id" type="I">
+      <constraint>REFERENCES cm_email_widget_emails(id)</constraint>
+    </field>
+    <field name="created" type="T" />
+    <field name="modified" type="T" />
+    <field name="email" type="C" size="128" />
+    <field name="type" type="C" size="32" />
+    <field name="primary_email" type="L" />
```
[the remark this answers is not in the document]
spartanicus: Correct. It's been removed.

app/AvailablePlugin/EmailWidget/Config/Schema/schema.xml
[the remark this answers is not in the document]
spartanicus: (i.e. Done.)

app/AvailablePlugin/EmailWidget/Controller/CoEmailWidgetsController.php
```diff
@@ -0,0 +1,114 @@
...
+   * Render the widget according to the requested user and current configuration.
+   *
+   * @since  COmanage Registry v4.1.0
+   * @param  Integer $id CO Services Widget ID
+   */
+  
+  public function display($id) {
+    $cfg = $this->CoEmailWidget->getConfig();
+    
+    // Pass the config so we know which div to overwrite
```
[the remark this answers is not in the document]
spartanicus: Done.

app/AvailablePlugin/EmailWidget/Model/EmailWidgetEmail.php
spartanicus: Done.

app/AvailablePlugin/EmailWidget/Controller/EmailWidgetEmailsController.php
```diff
@@ -0,0 +1,159 @@
...
+  public function gentoken() {
+    if(!empty($this->request->data['email'])
+       && !empty($this->request->data["configid"])) {
+      $this->CoEmailWidget->id = $this->request->data["configid"];
+
+      $email = $this->request->data['email'];
+      $primary = $this->request->data['primary'];
+      
+      $results = $this->EmailWidgetEmail->generateToken($email,$this->CoEmailWidget->field('default_type'),$primary);
+      if(!empty($results['id'])) { // XXX Ensure this test is adequate.
```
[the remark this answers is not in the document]
spartanicus: Yes - comment removed.

app/AvailablePlugin/EmailWidget/Controller/EmailWidgetEmailsController.php
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * Step 1 in adding an email address: 
+   * Generate and mail a token for email verification.
+   * Return the row id back for use in step two. 
+   *
+   * @since  COmanage Registry v4.1.0
+   */
+  public function gentoken() {
+    if(!empty($this->request->data['email'])
+       && !empty($this->request->data["configid"])) {
```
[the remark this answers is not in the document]
spartanicus: Good point: done.

app/AvailablePlugin/EmailWidget/Controller/EmailWidgetEmailsController.php
```diff
@@ -0,0 +1,159 @@
...
+    'CoMessageTemplate'
+  );
+
+  public function beforeFilter() {
+    parent::beforeFilter();
+
+    if(in_array($this->action, EmailWidgetEmailsController::$actions)) {
+      $this->Security->validatePost = false;
+      $this->Security->enabled = false;
+      $this->Security->csrfCheck = false;
```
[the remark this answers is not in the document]
spartanicus: Passing this along to @ioigoume 
Note: configid is now 'email_address_widet_id'

app/Lib/lang.php
spartanicus: This was actually inserted to represent an HTTP 500 error that could occur during the ajax call and bubbled back up to the browser. I've changed this to be more explicit:
`'er.500' =>         'Internal server error (500)'`

app/AvailablePlugin/EmailWidget/Config/Schema/schema.xml
```diff
@@ -0,0 +1,69 @@
...
+    <field name="co_message_template_id" type="I">
+      <constraint>REFERENCES cm_co_message_templates(id)</constraint>
+    </field>
+    
+    <index name="co_email_widgets_i1">
+      <col>co_dashboard_widget_id</col>
+      <unique />
+    </index>
+  </table>
+  <table name="email_widget_emails">
```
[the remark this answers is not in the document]
spartanicus: Agreed - the name has been changed as suggested. (The foreign key still needs adding, however.)

spartanicus:
This PR has been rebased against the latest develop.

app/AvailablePlugin/EmailWidget/Controller/EmailWidgetEmailsController.php
```diff
@@ -0,0 +1,159 @@
...
+   * or invoke the parent method.
+   * - precondition: A coid should be provided in the query string
+   *
+   * @since  COmanage Registry v4.1.0
+   * @return Integer The CO ID if found, or -1 if not
+   */
+
+  public function parseCOID($data = null) {
+    if($this->request->method() == "POST"
+       && isset($this->request->data["coid"])
```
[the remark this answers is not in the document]
spartanicus: This has been changed in Ioannis's back-end changes.

app/AvailablePlugin/EmailWidget/Controller/EmailWidgetEmailsController.php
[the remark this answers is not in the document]
spartanicus: This has been updated in Ioannis's back-end changes.

app/AvailablePlugin/EmailWidget/Controller/EmailWidgetEmailsController.php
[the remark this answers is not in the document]
spartanicus: This has been updated in Ioannis's back-end changes.

app/AvailablePlugin/EmailWidget/Controller/EmailWidgetEmailsController.php
```diff
@@ -0,0 +1,159 @@
...
+   * @since  COmanage Registry v4.1.0
+   */
+  public function verify() {
+    if(!empty($this->request->data['token']) &&
+       !empty($this->request->data['id']) &&
+       !empty($this->request->data['copersonid'])) {
+      $token = $this->request->data['token'];
+      $id = $this->request->data['id'];
+      $coPersonId = $this->request->data['copersonid'];
+      $outcome = $this->EmailWidgetEmail->verify($token,$id,$coPersonId);
```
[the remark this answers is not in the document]
spartanicus: Also updated in Ioannis's back-end changes.

app/AvailablePlugin/EmailAddressWidget/Model/EmailAddressWidgetVerification.php
```diff
@@ -0,0 +1,122 @@
...
+    $timeElapsed = time() - strtotime($this->rec['EmailAddressWidgetVerification']['created']);
+    $timeWindow = (int)$this->rec["CoEmailAddressWidget"]["verification_validity"] * 60;
+
+    if($timeElapsed > $timeWindow) {
+      // Delete the record and return
+      $this->delete($this->rec['EmailAddressWidgetVerification']['id']);
+      return "timeout"; // token timed out
+    }
+
+    // Create the new CO Person Email record
```
boshrin, line 101: We should also create a History Record so that an admin can see that the user added and verified the address via the self service widget.

app/Lib/lang.php
```diff
@@ -1060,6 +1060,7 @@
   ),
   
   // Errors
+  'er.500' =>         'Internal server error (500)',
```
boshrin, line 1063: If we're going to start localizing HTTP codes, they should probably be grouped together, eg `er.http.500`.

app/View/Elements/javascript.ctp
```diff
@@ -657,4 +657,8 @@
   var defaultConfirmOk = "<?php print _txt('op.ok'); ?>";
   var defaultConfirmCancel = "<?php print _txt('op.cancel'); ?>";
   var defaultConfirmTitle = "<?php print _txt('op.confirm'); ?>";
+  
+  // Define default text for session timeout and unknown errors
+  var defaultHttp500 = "<?php print _txt('er.500'); ?>";
+  var defaultHttp401 = "<?php print _txt('er.timeout'); ?>";
```
boshrin, line 663: Note the inconsistent use of error tags here, why isn't the second one `er.http.401`?

app/AvailablePlugin/EmailAddressWidget/Config/Schema/schema.xml
```diff
@@ -0,0 +1,80 @@
...
+      <constraint>REFERENCES cm_co_dashboard_widgets(id)</constraint>
+    </field>
+    <field name="created" type="T" />
+    <field name="modified" type="T" />
+
+    <index name="co_email_address_widgets_i1">
+      <col>co_dashboard_widget_id</col>
+      <unique />
+    </index>
+    <index name="co_email_address_widgets_i2">
```
boshrin, line 50: Do we select EmailAddress Widgets based on their message template? If not, we don't need this index.

app/AvailablePlugin/EmailAddressWidget/Config/Schema/schema.xml
```diff
@@ -0,0 +1,80 @@
...
+    </field>
+    <field name="created" type="T" />
+    <field name="modified" type="T" />
+
+    <index name="email_address_widget_verifications_i1">
+      <col>co_email_address_widget_id</col>
+    </index>
+    <index name="email_address_widget_verifications_i2">
+      <col>co_person_id</col>
+    </index>
```
boshrin, line 78: I'm not sure we need either of these indexes, but the one we actually do need (`token`, which is what `calculateImpliedCoId` uses) is missing.

boshrin (changes requested):

app/AvailablePlugin/EmailAddressWidget/Lib/lang.php
```diff
@@ -0,0 +1,86 @@
...
+// a different order than they were passed.
+
+$cm_email_address_widget_texts['en_US'] = array(
+  // Titles, per-controller
+  'ct.co_email_address_widgets.1'  => 'Self Service Email Widget',
+  'ct.co_email_address_widgets.pl' => 'Self Service Email Widgets',
+  
+  // Error
+  'er.emailaddresswidget.401' => 'Your session has timed out. Please log in.',
+  'er.emailaddresswidget.500' => 'There was an unknown error.',
```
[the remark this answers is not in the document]
spartanicus: This should be aligned with the general 500 error, I'd think -- or if this one is in use, perhaps we just make use of the general 500 error added to the core lang file.

app/View/Elements/javascript.ctp
spartanicus: For the plugin we may combine two messages - one that's generic ("Unauthorized (401)" with one that's specific "Your session has timed out. Please log in to continue." Or we can make the generic 401 contain both phrases: "Unauthorized (401). Your session has timed out. Please log in to continue." Would this (or something like this) be appropriate for all 401s?

app/AvailablePlugin/EmailAddressWidget/Lib/lang.php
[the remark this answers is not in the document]
spartanicus: This is done, but see notes below.

app/Lib/lang.php
spartanicus: See the following comment (question) about er.http.401.

app/View/Elements/javascript.ctp
spartanicus: I'm not sure, however, how useful the HTTP error code is to the end user. I think just leaving the end user message as 'er.timeout' is probably better. A developer can look under the hood to see the code returned. 

er.http.500 is different: it's a server error of some kind that just needs to be reported. I had originally told the end user "Unknown error" - but I think being more explicit here is good. (It's now "Internal server error (500)").

But I feel like an end user message for 401 should probably remain er.timeout: it's simple to understand and involves no tech speak in the end-user widget.

spartanicus: I was considering HTTP 404 as a counter example, but lots of places don't report the status code to end users anymore:

- https://www.amazon.com/askdfnnn
- https://www.nytimes.com/asdfslkjalskfjslkdfjslkj

Devs just look at the network response.

spartanicus: This has me thinking that "Unknown error" (or something more explicit) may still be better for a 500. Something like "The server threw an unknown error." may be better than reporting the HTTP status code to the end user. Or again, we could just make the error message a little more explicit: 

- "HTTP Error 500: The server threw an unknown error."

If we compromise on this, we could do something similar for 401:

- "HTTP Error 401: Your session has expired. Please login to continue."

If we want a compromise, how's that?

spartanicus: The two examples above, if used, would probably map to the er.http.500 and er.http.401 lang strings. 

But I'm still somewhat inclined to just skip the error codes altogether, and have er.unknown and er.timeout -- Or 'er.http.500.unknown' and 'er.http.401.timeout' to make the lang strings themselves more explicit about what they represent, but the strings would just be 

- "The server threw an unknown error."
- "Your session has expired. Please login to continue."

Thoughts?

app/AvailablePlugin/EmailAddressWidget/Config/Schema/schema.xml
```diff
@@ -0,0 +1,75 @@
...
+  While this schema mostly works with database prefixes, foreign key constraints
+  must be specified in raw SQL, which needs the prefixed table name.
+-->
+<schema version="0.3">
+  <table name="co_email_address_widgets">
+    <field name="id" type="I">
+      <key />
+      <autoincrement />
+    </field>
+    <field name="type" type="C" size="128" />
```
boshrin, line 33: This should be `varchar(32)` to align with the Extended Type definition.

app/AvailablePlugin/EmailAddressWidget/Controller/CoEmailAddressWidgetsController.php
```diff
@@ -0,0 +1,178 @@
...
+        return;
+      }
+
+      // I need to verify that the CO Person is part of the CO
+      $copersonid = $this->request->query['copersonid'];
+      if(!$this->Role->isCoPerson($copersonid, $this->cur_co["Co"]["id"])) {
+        $this->Api->restResultHeader(HttpStatusCodesEnum::HTTP_NOT_FOUND, _txt('er.cop.nf', array($copersonid)));
+        return;
+      }
+    }
```
boshrin, line 63: Since this logic is specific to `gentoken()`, it should be moved there.

app/AvailablePlugin/EmailAddressWidget/Controller/CoEmailAddressWidgetsController.php
```diff
@@ -0,0 +1,178 @@
...
+
+      // We have a valid result. Provide the row ID to the verification form,
+      // and send the token to the new email address for round-trip verification by the user.
+      $this->Api->restResultHeader(HttpStatusCodesEnum::HTTP_CREATED,  _txt('fd.created'));
+      $this->set('vv_token', $results['token']);
+      $this->CoEmailAddressWidget->send($this->request->query['email'],
+                                        $results['token'],
+                                        $this->CoEmailAddressWidget->field('co_message_template_id'));
+      return;
+    }
```
boshrin, line 94: Similarly this all belongs in `gentoken()`.

app/AvailablePlugin/EmailAddressWidget/Controller/CoEmailAddressWidgetsController.php
```diff
@@ -0,0 +1,178 @@
...
+   *
+   * @since  COmanage Registry v4.1.0
+   * @param  Integer $id CO Services Widget ID
+   */
+  
+  public function display($id) {
+    // We need only the CoPerson ID - with that we can look up the Email Addresses via 
+    // ajax against the API in the web client.
+    $coPersonId = $this->reqCoPersonId;
+    $this->set('vv_co_person_id', $coPersonId);
```
boshrin, line 123: Since `$coPersonId` isn't used anywhere else these should be merged into one line.

app/AvailablePlugin/EmailAddressWidget/Controller/EmailAddressWidgetVerificationsController.php
```diff
@@ -0,0 +1,139 @@
...
+        $this->Api->restResultHeader(HttpStatusCodesEnum::HTTP_NOT_ACCEPTABLE, _txt('er.emailaddresswidget.timeout'));
+        return false;
+      }
+      $this->set('email_address_id', $this->EmailAddressWidgetVerification->addEmailToPerson($token, $actorCoPersonId));
+
+      $this->Api->restResultHeader(HttpStatusCodesEnum::HTTP_CREATED,
+                                   _txt('rs.added-a3', array($rec['EmailAddressWidgetVerification']['email'])));
+    } catch (Exception $e) {
+      $this->Api->restResultHeader(HttpStatusCodesEnum::HTTP_BAD_REQUEST, $e->getMessage());
+    }
```
boshrin, line 100: All of this logic should be moved into the model, which can throw an appropriate error (if needed) that the controller can catch and use to return back via ApiComponent.

app/AvailablePlugin/EmailAddressWidget/Controller/EmailAddressWidgetVerificationsController.php
```diff
@@ -0,0 +1,139 @@
...
+
+    // Retrieve the Verification record and the configuration
+    $rec = $this->EmailAddressWidgetVerification->getRecordToVerify($token);
+    if(empty($rec)) {
+      $this->Api->restResultHeader(HttpStatusCodesEnum::HTTP_NOT_FOUND, _txt('er.token'));
+      return false;
+    }
+    // Check if the person owns the records and can proceed with the verification
+    $CoPerson = ClassRegistry::init('CoPerson');
+    $actorCoPersonId = $CoPerson->idForIdentifier($this->cur_co["Co"]["id"], $this->Session->read('Auth.User.username'));
```
boshrin, line 81: I'm pretty sure I commented on this before. Don't do this identifier map. Just read `Auth.User.co_person_id` from the session.

app/AvailablePlugin/EmailAddressWidget/Model/CoEmailAddressWidget.php
```diff
@@ -0,0 +1,151 @@
...
+    );
+
+    $this->EmailAddressWidgetVerification->clear();
+    if(!$this->EmailAddressWidgetVerification->save($fields)) {
+      throw new RuntimeException($this->validationErrors);
+    }
+
+    // Return the token to pass along via email
+    return array(
+      'token' => $token
```
boshrin, line 108: Why are we returning an array with a single key? Why don't we just return the token?

Also, either way this isn't consistent with the docblock signature above.

app/AvailablePlugin/EmailAddressWidget/Model/EmailAddressWidgetVerification.php
```diff
@@ -0,0 +1,193 @@
...
+  );
+
+  // Validation rules for table elements
+  public $validate = array(
+    'email' => array(
+      'content' => array(
+        'rule' => array('email'),
+        'required' => false,
+        'allowEmpty' => false,
+        'message' => 'Please enter a valid email address'
```
boshrin, line 55: Email can't be both not required and not allowed empty. In this case, `required` should probably be `true`.

Also, don't hardcode english error messages. I know there are some places that still do this, but we shouldn't be adding any more.

app/AvailablePlugin/EmailAddressWidget/Model/EmailAddressWidgetVerification.php
```diff
@@ -0,0 +1,193 @@
...
+                                       ActionEnum::CoPersonEditedManual,
+                                       _txt('pl.emailaddresswidget.rs.added-a', array($rec['EmailAddressWidgetVerification']['email'])));
+
+      // History Record for the verification
+      $CoPerson->HistoryRecord->record($rec['EmailAddressWidgetVerification']['co_person_id'],
+                                       null,
+                                       null,
+                                       $actorCoPersonId,
+                                       ActionEnum::EmailAddressVerified,
+                                       _txt('pl.emailaddresswidget.rs.mail.verified', array($rec['EmailAddressWidgetVerification']['email'])));
```
boshrin, line 136: I think we could collapse these into a single message: "Added Verified Email Address %1$s"

app/AvailablePlugin/EmailAddressWidget/Model/EmailAddressWidgetVerification.php
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * Get Verification Record Singleton
+   *
+   * @since  COmanage Registry v4.1.0
+   * @param  string $token     Token used for verification
+   * @return EmailAddressWidgetVerification Verification Record
+   */
+
+  public function getRecordToVerify($token) {
+    if(empty($this->_recordToVerify)) {
```
boshrin, line 155: As per my comment above, the verification steps should be consolidated from the controller into a single model call. Then we won't need to cache the record.

app/AvailablePlugin/EmailAddressWidget/webroot/js/email-panel.js
```diff
@@ -0,0 +1,236 @@
...
+      // basic front-end validation: does it contain '@' and '.'?
+      if(newEmailAddress.indexOf('@') == -1 || newEmailAddress.indexOf('.') == -1) {
+        this.newEmailInvalid = true;
+        this.newEmailInvalidClass = 'is-invalid';
+        this.newEmailErrorMessage = this.txt.errorInvalid;
+        return;
+      }
+
+      // URL: /donations/view/recent/2001
+      // Mapping: DonationsController->view('recent', '2001');
```
boshrin, line 68: DonationsController?

app/AvailablePlugin/EmailAddressWidget/webroot/js/email-panel.js
```diff
@@ -0,0 +1,236 @@
...
+      // XXX Is the copersonid part of the CO?
```
boshrin, line 69: Is it?

boshrin (changes requested):

boshrin (approved):

PR #404, opened 2022-08-19
CO-2383_Cannot_create_LDAP_Provisioner_table_with_MySQL_8.0.1_or_later

ioigoume:
Wrap columns into back ticks for MySQL database creation


boshrin:
This works for me. Would be good to get confirmation from @skoranda .

PR #408, opened 2022-08-23
Fix nested AND group reconcile logic (CO-2491)

tomzeller:
Attempt to fix typo in nested AND group reconcile logic

### DIFF
--- a/app/Model/CoGroup.php
+++ b/app/Model/CoGroup.php
@@ -715,7 +715,7 @@ class CoGroup extends AppModel {
         continue;
       }
       
-      if(!isset($nMembers[ $t['CoGroupMember']['co_group_nesting_id'] ][ $t['CoGroupMember']['co_person_id'] ])) {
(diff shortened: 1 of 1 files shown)


boshrin:
I'm not sure I understand this patch.

`$t['CoGroupMember']['co_group_nesting_id']` is the CO Group Nesting ID that created the current membership we are examining.

`$t['CoGroupMember']['co_person_id']` is the CO Person ID we are currently examining.

`$nMembers` is the set of current memberships, keyed in CO Group Nesting ID and then CO Person ID. So `!isset(...)` says "if the specified CO Person is not a member of the group due to this nesting, remove the nested membership. Won't flipping the negation cause the membership to be removed if the person _is_ currently a member? They'd get added back in in the next `foreach`, but that would basically remove and then add every member during each call to `reconcile`, which would generate a lot of noise.

boshrin:
Obviated by c6d0c80160. Due to the introduction of boolean AND, `syncNestedMembership()` basically needs to be called all the time. This is less efficient, but will suffice for v4.

PR #409, opened 2022-08-23
Co 2492

tomzeller:
Ignore suspended nested groups.


boshrin:
While this works, it would be less code to use Containable to pull the group of interest. ie, change line 918 to something like `$args['contain'] = array('CoGroup')`. Then within the foreach loop you can simply examine the group's status, which will be under the nesting.

boshrin:
Obviated by c6d0c80160. This check also needed to be done in `CoGroup.php`.

PR #410, opened 2022-08-30
CO-2484_Elector_Data_Filter

ioigoume:


app/AvailablePlugin/ElectorDataFilter/View/ElectorDataFilterPrecedences/add.ctp
```diff
@@ -0,0 +1,76 @@
+<?php
```
boshrin, line 1: Are these view files actually files or symlinks? (They should just be symlinks to the Standard view directory.)

app/AvailablePlugin/ElectorDataFilter/View/Elements/mvep.ctp
```diff
@@ -0,0 +1,157 @@
+<?php
```
boshrin, line 1: Is this file supposed to be plugin specific?

app/Model/Address.php
```diff
@@ -28,6 +28,9 @@
 class Address extends AppModel {
   // Define class name for cake
   public $name = "Address";
+
```
boshrin, line 31: Per the comment above, I'm not sure we want to keep these.

boshrin (changes requested):

PR #419, opened 2022-09-25
Review_fixes

ioigoume:
Review fixes


app/AvailablePlugin/EmailAddressWidget/Controller/EmailAddressWidgetVerificationsController.php
```diff
@@ -72,25 +72,29 @@ public function verify($token) {
 
     // Retrieve the Verification record and the configuration
     $rec = $this->EmailAddressWidgetVerification->getRecordToVerify($token);
+    if(empty($rec)) {
+      $this->Api->restResultHeader(HttpStatusCodesEnum::HTTP_NOT_FOUND, _txt('er.notfound-b', array($token)));
```
boshrin, line 76: If this error is going to be displayed to the user, I'm not sure the actual token is a meaningful message. Something like "Token not found" might be better.

spartanicus: I agree with Benn, and I think the most meaningful message is probably "Invalid token". I’m not sure an end user would understand what “not found” means in this context. But they will understand a simple message that says that the token is no good. So, "Invalid token" would be my suggestion.

PR #423, opened 2022-09-28
CO-2464_Override_Full_Base_URL

ioigoume:
Override fullBaseUrl using env variables

### DIFF
--- a/app/Config/bootstrap.php
+++ b/app/Config/bootstrap.php
@@ -48,6 +48,16 @@ define('LOCAL', ROOT . DS . 'local' . DS);
  *
  */
 
+/**
+ * Full URL prefix
+ * Get and set the fullBaseUrl using an environmental variable
+ * e.g. Apache server configuration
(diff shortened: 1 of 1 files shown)


app/Config/bootstrap.php
```diff
@@ -48,6 +48,18 @@
  *
  */
 
+/**
+ * Full URL prefix
+ * Get and set the fullBaseUrl using an environmental variable
+ * e.g. Apache server configuration
+ * SetEnv COMANAGE_REGISTRY_FULL_BASE_URL https://example.com
+ */
+if (!defined('CMG_FULL_BASE_URL')
```
boshrin, line 57: Do we really need to defined `CMG_FULL_BASE_URL`? If someone wants to define the URL via Cake configuration, they can just set `App.fullBaseUrl` directly (and not pass in `$COMANAGE_REGISTRY_FULL_BASE_URL`).

PR #424, opened 2022-10-01
CO-2532_Petition_Attributes_Introductory_Text

ioigoume:
Add introductory text on top of Petition Form


app/Config/Schema/schema.xml
```diff
@@ -1109,6 +1109,7 @@
     </field>
     <field name="introduction_text" type="X" size="4000" />
     <field name="conclusion_text" type="X" size="4000" />
+    <field name="introduction_text_pf" type="X" size="4000" />
```
boshrin, line 1112: Maybe use the step name so it's clearer? `introduction_text_petitioner_attributes`, or if that's too long `introduction_text_pa`?

boshrin (commented):
Should there be an update to `CoEnrollmentFlows/fields.inc` to configure the setting?

PR #427, opened 2022-10-04
CO-2533_Filter_CO_Group_Members_by_CO_Person

ioigoume:
filter CO Group Members by CO Person Id


boshrin (changes requested):
If I provide both `cogroupid` and `copersonid`, it looks like `cogroupid` is checked first, and then only if no results are found `copersonid` is tried. This could create inconsistent behavior depending on whether or not `cogroupid` returns any results.

If both are provided, we should treat it as an AND query - ie only return CoGroupMemberships where both `cogroupid` and `copersonid` match.

We also need to check that the authorization calculation is done correctly regardless of which parameter is provided (ie: that the requester is a privileged API user for whichever arguments are provided) if you didn't already do that.

PR #434, opened 2022-11-02
CO-1637_Disable_Non-Admin_CO_Group_Creation

ioigoume:
Disable CO Group creation from non-admins

### DIFF
--- a/app/Config/Schema/schema.xml
+++ b/app/Config/Schema/schema.xml
@@ -3087,6 +3087,7 @@
     <field name="enable_nsf_demo" type="L" />
     <field name="disable_expiration" type="L" />
     <field name="disable_ois_sync" type="L" />
+    <field name="group_create_admin_only" type="L" />
     <field name="group_validity_sync_window" type="I" />
     <field name="garbage_collection_interval" type="I" />
     <field name="enable_normalization" type="L" />
--- a/app/Controller/CoGroupsController.php
+++ b/app/Controller/CoGroupsController.php
@@ -434,8 +434,15 @@ class CoGroupsController extends StandardController {
     // Determine what operations this user can perform
     
     // Add a new Group?
-    $p['add'] = ($roles['cmadmin'] || $roles['coadmin'] || $roles['comember']);
-    
+    $p['add'] = $roles['cmadmin']
+                || $roles['coadmin']
+                || $roles['couadmin']
+                || ($this->CoGroup->Co->CoSetting->allowGroupCreationByNonAdmins($this->cur_co['Co']['id'])
+                    && $roles['comember']); // XXX CO-1637
+
+    // Manage the Group memberships
+    $p['groupmanage'] = ($roles['cmadmin'] || $roles['coadmin'] || $roles['comember']);
+
     // Create an admin Group?
     $p['admin'] = ($roles['cmadmin'] || $roles['coadmin']);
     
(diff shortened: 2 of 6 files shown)


app/Config/Schema/schema.xml
```diff
@@ -3087,6 +3087,7 @@
     <field name="enable_nsf_demo" type="L" />
     <field name="disable_expiration" type="L" />
     <field name="disable_ois_sync" type="L" />
+    <field name="disable_nonadmins_grp_crt" type="L" />
```
boshrin, line 3090: I realize we're limited in how long the column name can be, but I wonder if we can somehow be more descriptive. Maybe `group_create_admin_only`?

app/Controller/CoGroupsController.php
```diff
@@ -439,7 +439,14 @@ function isAuthorized() {
     
     // Add a new Group?
     $p['add'] = ($roles['cmadmin'] || $roles['coadmin'] || $roles['comember']);
-    
+    // XXX CO-1637
```
boshrin, line 442: XXX

Also, it would be clearer to do something like

```
$p['add'] = $roles['cmadmin'] || $roles['coadmin'] || $roles[`couadmin'] 
          || ($this->...->CoSetting->nonAdminsGroupCreate(...) && $roles['comember']);
```

app/Controller/CoGroupsController.php
```diff
@@ -463,7 +470,10 @@ function isAuthorized() {
     // View all existing Groups?
     $p['index'] = ($roles['cmadmin'] || $roles['coadmin'] || $roles['comember']);
     $p['search'] = $p['index'];
-    
+    if(!$this->CoGroup->Co->CoSetting->nonAdminsGroupCreate($this->cur_co['Co']['id'])) {
+      $p['search'] = ($roles['cmadmin'] || $roles['coadmin'] || $roles['couadmin']);
```
boshrin, line 474: Why is the search permission being changed? Isn't this RFE to only change the add permission?

app/Model/CoSetting.php
```diff
@@ -474,7 +480,21 @@ public function getThemeId($coId) {
...
+
+  /**
+   * Determine if Non-Admin Users can create Groups for the specified CO.
+   *
+   * @since  COmanage Registry v4.1.0
+   * @param  integer $coId CO ID
+   * @return boolean True if enabled, false otherwise
+   */
+
+  public function nonAdminsGroupCreate($coId) {
```
boshrin, line 492: We could be a bit wordier/clearer with this function name. Maybe something like `allowGroupCreationByNonAdmins()`?

app/Model/CoSetting.php
```diff
@@ -585,4 +605,33 @@ public function _ug400() {
...
   }
+
+  /**
+   * Perform CoSetting model upgrade steps for version 4.1.0.
+   * This function should only be called by UpgradeVersionShell.
+   *
+   * @since  COmanage Registry v4.1.0
+   */
+
+  public function _ug410() {
```
boshrin, line 616: I don't think we need an upgradeShell function (which doesn't look like it's called anyway) since `CoSetting::lookupValue()` will use the default value (via `$defaultSettings`) if no value is found.

boshrin (changes requested):

app/Controller/CoGroupsController.php
boshrin: Haha, stop pointing me to my own comments.

I'm actually not sure why that comment went into CO-1637, CO-1606 would make more sense. I'm going to move it there.

PR #441, opened 2022-11-06
System of Record Label fixes (CO-2547)

spartanicus:
Make System of Record Label field not required on Enrollment Flows, add description to the field, and move the field to a better location in the form for both Enrollment Flows and OrgID Sources.


skoranda:
The committed code runs as I would expect.

PR #449, opened 2023-01-03
CO-2380_Errors_when_deleting_non-empty_CO

ioigoume:
Unload unnecessary associations during a CO delete

### DIFF
--- a/app/AvailablePlugin/ApiSource/Model/ApiSource.php
+++ b/app/AvailablePlugin/ApiSource/Model/ApiSource.php
@@ -30,7 +30,10 @@ class ApiSource extends AppModel {
   public $cmPluginType = array("orgidsource", "job");
   
   // Document foreign keys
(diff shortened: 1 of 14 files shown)


boshrin:
In general, adding `dependent => true` is not the correct behavior for most of the models here. eg: Deleting a CoMessageTemplate should not delete the Authenticator that points to it. Deleting a CO Person should not delete History Records where that CO Person was the Actor (this is why Expunge rewrites those records). Deleting a Pipeline should not delete an OrgIdentitySource that was configured to point to it.

The more fundamental problem in that Cake's dependency processing isn't that smart, so when you delete a CO it may or may not correctly delete things in the right order. This is why there's already a bunch of special code for CO deletion -- that's where changes probably need to be made to make sure the records get deleted in the right order.

boshrin:
Please change the base branch back to `develop`.

PR #458, opened 2023-01-18
CO-2563_CO_Person_save_takes_too_long_for_deployments_with_many_COUs

ioigoume:


boshrin:
This looks OK but it'll need to be moved to `develop` instead of `hotfix-4.1.x`.

PR #465, opened 2023-02-07
CO-2600_My_population_search_filter_double-encoding_@_signs

ioigoume:
Fix forward slash when used as part of a search term.

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -1635,6 +1635,68 @@ class AppController extends Controller {
   }
   
   /**
+   * Redirects to given $url, after turning off $this->autoRender.
+   * Script execution is halted after the redirect.
+   *
+   * @param string|array $url A string or array-based URL pointing to another location within the app,
+   *     or an absolute URL
+   * @param int|array|null|string $status HTTP status code (eg: 301). Defaults to 302 when null is passed.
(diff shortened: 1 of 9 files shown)


spartanicus:
Hi, @ioigoume - this fixes the problem with filters with the "@" sign. Those work again. The original issue of encoding the "/" sign is not working for me. 

To reproduce: 
1. Add a new name to a CO Person with the Given Name "A name / with a / slash"
2. Use the My Population filters on Given Name for "A name / with"

My result is "No results found". 

The path in my browser looks like this:
`/registry/co_people/index/co:2/search.givenName:A%20name%20%252F%20with/op:search`

Note the extra characters in the middle, where the slash is represented: `%252F` (%25 = '%' and %2F = '/'). So this looks like double encoding. 

Peculiarly, if I remove the `2F` so that the url looks like
`/registry/co_people/index/co:2/search.givenName:A%20name%20%25%20with/op:search`
(which would be a '%' sign) the filter works, and I get the Person I expect in the filtered listing.

Putting the encoded slash `%2F` in the search rather than `%25` breaks because the browser reads it as a "/". It's interesting that including the %25 makes this search work.

spartanicus:
This looks great. Merging.

PR #467, opened 2023-02-19
CO-2321_Create_MaxRetry_counter_for_queue_LDAP_provisioning_jobs

ioigoume:


app/Config/Schema/schema.xml
```diff
@@ -1250,6 +1252,39 @@
     </index>
   </table>
   
+  <table name="co_provisioning_counts">
```
boshrin, line 1255: I'm guessing this gets its own table because there's no existing equivalent of `co_jobs`? It'd be nice if both mechanisms used the same pattern (eg: `co_job_counts` to match `co_provisioning_counts`) but maybe we should wait until PE to fix that since the provisioning implementation is changing quite a bit anyway. We could maybe leverage the new `provisioning_history_records` table somehow. Although we do currently have `co_provisioning_exports`, maybe that could be used somehow?

app/Controller/CoJobsController.php
```diff
@@ -130,11 +130,21 @@ static function ($model) use ($plugin_name) {
...
-        // Request the page auto-refresh
+      } elseif($this->action == 'view') {
 
-        $this->set('vv_refresh_interval', 15);
+        if(in_array($this->viewVars['co_jobs'][0]['CoJob']['status'], array(JobStatusEnum::InProgress, JobStatusEnum::Queued))) {
+          // Request the page auto-refresh
+          $this->set('vv_refresh_interval', 15);
+        }
+
+        // Get the provisioing Target
```
boshrin, line 140: typo

app/Lib/lang.php
```diff
@@ -1255,6 +1256,7 @@
   'er.pooling' =>     'Org Identity Pooling not supported in this context',
   // er.prov is a javascript string and so cannot take a parameter
   'er.prov' =>        'Provisioning failed: ',
+  'er.prov.job.count' => 'ProvisionigCount record failed to save',
```
boshrin, line 1259: typo

app/Lib/lang.php
```diff
@@ -1670,6 +1672,10 @@
   'fd.ip' =>          'IP Address',
   'fd.job.complete_time' => 'Finished',
   'fd.job.finish_summary' => 'Finish Summary',
+  'fd.job.max_retry' => 'Maximum Retry',
+  'fd.job.max_retry.desc' => 'If the job fails, it will automatically be rescheduled for <bold>"%1$s" number</bold> of times. (To stop retrying, simply cancel this job.)',
```
boshrin, line 1676: I think we could simplify this to "it will automatically be rescheduled up to %1$s time(s)".

app/Lib/lang.php
```diff
@@ -1831,6 +1837,7 @@
   'fd.plugin.warn' => 'Once a new %1$s has been created, the Plugin cannot be changed',
   'fd.prov.group' =>  'Provisioning Group',
   'fd.prov.group.desc' => 'If set, only provision members of the specified group to this target',
+  'fd.prov.max_retry.desc' => 'After the provisioning fails it will automatically retry for XX times. If set to 0 (zero) it will retry for infinite times',
```
boshrin, line 1840: instead of XX this should say something like "... retry up to the specified number of times. If set to 0 (zero) it will retry indefinitely"

app/Model/CoJob.php
```diff
@@ -417,7 +463,32 @@ public function register($coId,
       // In order to present the error via the UI, we still want to insert a job record,
       // so we don't rollback.
     }
-    
+
+    // XXX Currently we support max retries only for CoreJob.Provision types
```
boshrin, line 467: Thinking about this some more, why should we special case this to Provision jobs? The max_retry parameter is on the CoJob, any type of Job could have a max retry count (even if the UI doesn't currently support setting that). This would also solve the "count" inconsistency I commented on above, we would simply track the count in co_jobs (like you already have).

boshrin (changes requested):

app/Lib/lang.php
ioigoume: done

app/Lib/lang.php
ioigoume: done

app/Lib/lang.php
ioigoume: done

app/Model/CoJob.php
ioigoume: the reason i used the `ProvisioningCount` model is because i followed the CoExpirationCount paradigm.
The max_retry is the configuration. We need a way to track the actual counts and link this counts back to the model. For the Expiration Policies there is a similar field the max_count.

PR #469, opened 2023-02-21
CO-2603_Normalizations_run_after_the_validations

ioigoume:
Set NormalizationBehavior beforeValidate instead of beforeSave callback


boshrin:
Pending discussion on if this should go into hotfix or stay on develop.

boshrin:
Well, the next release is 4.2.0 so develop it is...

PR #474, opened 2023-03-08
CO-2522_Filter_group_memberships_based_on_status

ioigoume:
Filter Group members based on member and owner status


boshrin:
Per the comment in CO-2522, should this PR be closed?

PR #475, opened 2023-03-11
CO-2331_Ensure_regular_groups_added_via_an_API_call_appear_in_the_Regular_Groups_listing

ioigoume:
Add default group type to regular groups created through the API

### DIFF
--- a/app/Model/CoGroup.php
+++ b/app/Model/CoGroup.php
@@ -495,6 +495,20 @@ class CoGroup extends AppModel {
         }
       }
     }
+
+    // If the group type is missing add the default Standard type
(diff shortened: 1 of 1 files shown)


spartanicus:
Hi, @ioigoume - this definitely works to add the group_type and that's necessary -- but that's not enough to fully fix the filtering issue: we also need to supply "auto" = f (false) for groups added this way if we want the existing filters to work properly.

This code alone will not make groups added via the API appear in the "Regular Groups" listing. But if you set auto to "f" in the database they will appear. So, I think we just need to check for that in the same way and set it for an API group add.

PR #479, opened 2023-03-22
CO-2615_IdentifierEnroller_plugin_can_be_called_multiple_times_using_browser_back_button

ioigoume:
Skip identifier enroller for existing types

### DIFF
--- a/app/AvailablePlugin/IdentifierEnroller/Controller/IdentifierEnrollerCoPetitionsController.php
+++ b/app/AvailablePlugin/IdentifierEnroller/Controller/IdentifierEnrollerCoPetitionsController.php
@@ -59,11 +59,32 @@ class IdentifierEnrollerCoPetitionsController extends CoPetitionsController {
       $this->redirect($onFinish);
     }
 
+    // Get the CO Person ID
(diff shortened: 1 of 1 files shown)


app/AvailablePlugin/IdentifierEnroller/Controller/IdentifierEnrollerCoPetitionsController.php
```diff
@@ -59,12 +59,37 @@ protected function execute_plugin_collectIdentifier($id, $onFinish) {
...
+    $coPersonId = $this->CoPetition->field('enrollee_co_person_id', array('CoPetition.id' => $id));
+
+    // Get the CoPerson Identifiers.
+    // If we already have a type of identifier that matches the one controlled by the plugin instance
+    // skip
+    $args = array();
+    $args['joins'][0]['table'] = 'co_people';
+    $args['joins'][0]['alias'] = 'CoPerson';
+    $args['joins'][0]['type'] = 'INNER';
+    $args['joins'][0]['conditions'][0] = 'CoPerson.id=Identifier.co_person_id';
```
boshrin, line 72: Care needs to be taken when joining tables since this can bypass ChangelogBehavior and return deleted rows.

But... what does this join do in the first place? Why not just find Identifiers where `Identifier.co_person_id = $coPersonId`?

PR #481, opened 2023-03-24
Support Core API index limit up to 1000 (CO-2624)

skoranda:


boshrin:
This looks ok, but @skoranda can you update the base branch to `develop`? The next release will be 4.2.0.

PR #492, opened 2023-05-04
CO-2641_Petition_Enrollment_Flow_Link_Broken_For_Archived_Flows

ioigoume:
Handle deleted and archived enrollment flows rendering in petition views


boshrin:
I think we're going to need a more descriptive warning, or else people might flip out that their Enrollment Flow has been "deleted", something like the equivalent of `fd.pt.archived` (which renders on the petition attributes). eg: "The configuration of this Enrollment Flow changed after the creation of this Petition."

PR #495, opened 2023-05-14
CO-2637_Email_for_petitions_Pending_Approval_should_be_configurable

ioigoume:
Approver Notifications Message Templates

### DIFF
--- a/app/Controller/CoEnrollmentFlowsController.php
+++ b/app/Controller/CoEnrollmentFlowsController.php
@@ -159,7 +159,7 @@ class CoEnrollmentFlowsController extends StandardController {
       $args['conditions']['status'] = SuspendableStatusEnum::Active;
       $args['conditions']['context'] = array(
         MessageTemplateEnum::EnrollmentApproval,
-        MessageTemplateEnum::ApproverNotification,
+        MessageTemplateEnum::EnrollmentApprover,
         MessageTemplateEnum::EnrollmentFinalization,
         MessageTemplateEnum::EnrollmentVerification
       );
--- a/app/Controller/CoEnrollmentFlowsController.php
+++ b/app/Controller/CoEnrollmentFlowsController.php
@@ -159,6 +159,7 @@ class CoEnrollmentFlowsController extends StandardController {
       $args['conditions']['status'] = SuspendableStatusEnum::Active;
       $args['conditions']['context'] = array(
         MessageTemplateEnum::EnrollmentApproval,
+        MessageTemplateEnum::ApproverNotification,
         MessageTemplateEnum::EnrollmentFinalization,
         MessageTemplateEnum::EnrollmentVerification
       );
--- a/app/Model/CoMessageTemplate.php
+++ b/app/Model/CoMessageTemplate.php
@@ -90,7 +90,7 @@ class CoMessageTemplate extends AppModel {
     ),
     'context' => array(
       'rule' => array('inList', array(MessageTemplateEnum::Authenticator,
-                                      MessageTemplateEnum::ApproverNotification,
+                                      MessageTemplateEnum::EnrollmentApprover,
                                       MessageTemplateEnum::EnrollmentApproval,
                                       MessageTemplateEnum::EnrollmentFinalization,
                                       MessageTemplateEnum::EnrollmentVerification,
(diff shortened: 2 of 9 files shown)


app/Controller/CoEnrollmentFlowsController.php
```diff
@@ -159,6 +159,7 @@ function beforeRender() {
       $args['conditions']['status'] = SuspendableStatusEnum::Active;
       $args['conditions']['context'] = array(
         MessageTemplateEnum::EnrollmentApproval,
+        MessageTemplateEnum::ApproverNotification,
```
boshrin, line 162: Do we need a new context for this? It's still an Approval related email, we could probably just use `EnrollmentApproval`. If we do want a new context, it should follow the existing naming pattern, eg `EnrollmentApprover`.

app/Model/CoMessageTemplate.php
```diff
@@ -295,25 +300,31 @@ public function templateSend($id, $toEmail, $substituteWith = array()) {
...
+      'TITLE'                => $substituteWith['TITLE'] ?? "Mr",
+      'VALID_FROM'           => $substituteWith['VALID_FROM'] ?? "YYYY/MM/DD",
+      'VALID_THROUGH'        => $substituteWith['VALID_THROUGH'] ?? "YYYY/MM/DD",
+      'TOKEN'                => $substituteWith['TOKEN'] ?? "123456",
+      'ENROLLMENT_NAME'      => $substituteWith['ENROLLMENT_NAME'] ?? "Join vo.example.org COU",
+      'ORIG_PETITION_STATUS' => $substituteWith['ORIG_PETITION_STATUS'] ?? "Created",
+      'NEW_PETITION_STATUS'  => $substituteWith['NEW_PETITION_STATUS'] ?? "Pending Approval",
+      'ENROLLEE_NAME'        => $substituteWith['ENROLLEE_NAME'] ?? "John Doe",
+      'SOURCE_URL'           => $substituteWith['SOURCE_URL'] ?? "http://example.org/registry/petition/view/1",
+      'NOTIFICATION_URL'     => $substituteWith['NOTIFICATION_URL'] ?? "http://example.org/registry/co_notifications/view/1",
```
boshrin, line 327: I'm not sure where these default values came from (I guess I missed the PR that added them) but we shouldn't be using these defaults as they might confuse admins who receive them. If we need a default text, it should just be "(unknown)" or "(?)" to indicate that something went wrong.

app/Model/CoPetition.php
```diff
@@ -2437,11 +2437,6 @@ public function sendApprovalNotification($id, $actorCoPersonId, $action='approva
     
     $args = array();
     $args['conditions']['CoPetition.id'] = $id;
-/*    $args['contain']['CoEnrollmentFlow'] = array(
-      'CoEnrollmentFlowAppMessageTemplate',
-      'CoEnrollmentFlowDenMessageTemplate',
-      'CoEnrollmentFlowFinMessageTemplate'
-    );*/
```
boshrin, line 2444: Any reason to commit a commented out block? If so there should be a note.

boshrin (changes requested):

app/Model/CoMessageTemplate.php
ioigoume: These default values are used only when testing a template. They do not interfere with the rest of the flow.

ioigoume: These default values are used only when testing a template. They do not interfere with the rest of the flow. The default, dummy value is needed because when testing the template there are no meaningful data available.

app/Model/CoPetition.php
ioigoume: The latest commit does not include this block of code

app/Controller/CoEnrollmentFlowsController.php
ioigoume: I renamed the enum variable from ApproverNotification to EnrollmentApprover. The Approver template has its own foreign key. As a result i followed the pattern already applied for the other types of Templates. As far as i am concerned and it is more clear to keep them separate.

app/Model/CoMessageTemplate.php
boshrin: ok. the default values for testing should be defined somewhere closer to the testing code then, but that's not a topic for this PR.

boshrin (approved):

PR #501, opened 2023-05-30
Password Dashboard Widget (CO-2648)

spartanicus:
There is still clean up to be done in the UI to show successful PUT responses from the API - but this is largely complete. The PR is being published so consideration can be given to changes to the Password Authenticator plugin -- in particular, allowing the plugin to handle encrypting incoming passwords using the API.


spartanicus:
Resolved by other PR that included this code.

PR #502, opened 2023-06-04
CO-2655_Add_empty_option_to_attribute_forms_during_enrollment

ioigoume:


app/Model/CoPetition.php
```diff
@@ -2102,6 +2102,7 @@ public function saveAttributes($id, $enrollmentFlowId, $requestData, $petitioner
       
       if($this->EnrolleeOrgIdentity->saveAssociated($orgData, array("validate" => false,
                                                                     "atomic" => true,
+                                                                    "petition" => true,
```
boshrin, line 2105: `petition` as a key word doesn't convey what the flag does. On top of that, we've disabled validation (`validate` => `false`) but then `petition` appears to trigger some sort of validation rule, which is confusing. On top of that, this change is only to allow an empty COU, but we've modified the options for saving all models (OrgIdentity, CoPerson, and CoPersonRole).

With the caveat that the CoPetition code is already way too complex (but that we're not going to do a heavy refactoring until PE), there are existing places that probably make more sense to add the empty check. The most appropriate is probably `CoPetition::adjustValidationRules()`, which already changes Cake's validation rules on `CoPersonRole::affiliation`.

Note that the actual validation is triggered before these saves, via the calls to `validateModel()` a few lines earlier.

ioigoume: @boshrin i removed the flag and moved the field validation inside the `validateModel()`

PR #511, opened 2023-06-28
CO-2652_Deletion_of_Extended_Types_for_Affiliation_causes_EnvSource_failure

ioigoume:


app/Plugin/EnvSource/Config/Schema/schema.xml
```diff
@@ -39,6 +39,7 @@
     </field>
     <field name="duplicate_mode" type="C" size="2" />
     <field name="redirect_on_duplicate" type="C" size="1024" />
+    <field name="default_affiliation" type="C" size="32" />
```
boshrin, line 42: The problem of setting a default affiliation is general, not specific to EnvSource. PE is probably going to address this in the Pipeline configuration, similar to the _New CO Person Status_ configuration that already exists.

ioigoume: @boshrin we are not deleting an affiliation which is in use but one that we hard coded around the framework as the default one, this is the affiliation `member`. See EnvSourceBackend line 235.
I could disable the ability of the administrator to delete the member affiliation.

PR #515, opened 2023-07-03
CO-2651_COU_Sorting_Not_Working_When_Sorting_By_Parent_COU

ioigoume:


app/Model/Cou.php
```diff
@@ -153,15 +153,38 @@ public function afterSave($created, $options = Array()) {
...
    * Obtain all COUs within a specified CO.
    *
-   * @since  COmanage Registry v0.4
-   * @param  integer CO ID
-   * @param  string Format, one of "names", "ids", or "hash" of id => name
+   * @param  Integer       $coId          CO ID
+   * @param  String        $format        Format, one of "names", "ids", or "hash" of id => name
+   * @param  Null|Boolean  $withChildren  If null retrieve all COUs.
+   *                                      If true retrieve all COUs with children.
+   *                                      If false retrieve all COUs with no Children
```
boshrin, line 160: Accepting a mixed null|boolean makes it easier down the road to confuse false and null. Also, it's not immediately clear what the difference is between these modes, ie: shouldn't "all COUs" include all child COUs?

I think maybe what you're trying to do here is offer the following options:

1. All COUs as a flat list
2. All COUs as a hierarchy
3. Only top level COUs

Although I'm not sure hierarchy actually returns a full tree, it seems to only pull the first level?

ioigoume: i think the name of the new parameter caused confusion. What i am trying to do is add a new filtering option. The option is COUs that have childers, that do not have children or we do not care and fetch all COUS. I renamed the attribute from `withChildren` to `isParent`

PR #520, opened 2023-07-23
CO-1407_Copy_Configuration_To_JSON

ioigoume:


app/Model/Identifier.php
```diff
@@ -129,7 +129,7 @@ class Identifier extends AppModel {
                                                  IdentifierEnum::StudentSOR,
                                                  IdentifierEnum::UID))),
         'required' => true,
-        'allowEmpty' => false
+        'allowEmpty' => true
       )
```
boshrin, line 133: Is this right? Why would we make this field required but then allow it to be empty?

ioigoume: Yes this is intentional. We need to be able to save an empty value.
```php
$identifier = '';
```

ioigoume: @boshrin 
i reverted the change.

PR #521, opened 2023-07-24
make all .co-info-topbox blocks consistent and move away from jQuery UI classes (CO-2672)

spartanicus:
This PR is very broad (touches many files) but is largely a mechanical change to refactor and make consistent all .co-info-topbox information blocks. It improves both the layout and the styling, and the PR establishes more appropriate markup when the underlying message itself has more complicated, block-level markup.


ioigoume (approved):
LGTM

PR #525, opened 2023-07-31
CO-2677

shaynasings:
…es as an authenticator for a CO Person


app/AvailablePlugin/PrivacyIdeaAuthenticator/Controller/PaperTokensController.php
```diff
@@ -0,0 +1,198 @@
...
+    }
+  }
+
+  /**
+   * Callback before other controller methods are invoked or views are rendered.
+   *
+   * @since  COmanage Registry v4.3.0
+   */
+
+  public function beforeFilter() {
```
ioigoume, line 89: Is there a reason why not to use the recommended CAKEPHP approach? if you wish to use a Model in your controller: https://book.cakephp.org/2/en/controllers.html#components-helpers-and-uses

e.g.
```
  public $uses = array(
    'EmailAddressWidget.CoEmailAddressWidget',
    'EmailAddress',
    'CO',
    'CoMessageTemplate'
  );
  ```
when using `$uses` the first element of the array should always be this controller's model.

File: app/AvailablePlugin/PrivacyIdeaAuthenticator/Model/PrivacyIdea.php [hunk 2]
```diff
@@ -188,15 +198,31 @@ public function createToken($privacyIdeaAuthenticator, $coPersonId) {
...
-    
-    // We don't persist the QR Data, but we do need to return it for rendering
-    $token['qr_data'] = $jresponse->detail->googleurl->img;
-    
+
+    switch ($token_type) {
+      case PrivacyIDEATokenTypeEnum::TOTP:
+        $token['confirmed'] = false;
+        $TotpToken = new TotpToken();
+        $TotpToken->save($token);
```
ioigoume, line 207: What will happen if save fails?

app/AvailablePlugin/PrivacyIdeaAuthenticator/Model/PrivacyIdea.php
```diff
@@ -188,15 +198,31 @@ public function createToken($privacyIdeaAuthenticator, $coPersonId) {
...
+      case PrivacyIDEATokenTypeEnum::Paper:
+        $PaperToken = new PaperToken();
+        $PaperToken->save($token);
+
+        // We don't persist the codes themselves but need to present them to the user for copying/printing
+        $token['otps'] = (array)$jresponse->detail->otps;
+        break;
+    }
+
+    if(!$jresponse->result->status) {
```
ioigoume, line 222: this is probably not needed. The same check exists at line 193

shaynasings:
rebased again latest develop

PR #527, opened 2023-08-03
CO-2671_EnvSource_doesn_t_with_multi-valued_email_attribute

ioigoume:


boshrin:
See comment on [CO-2671](https://todos.internet2.edu/browse/CO-2671).

app/Plugin/EnvSource/Config/Schema/schema.xml
```diff
@@ -39,6 +39,7 @@
     </field>
     <field name="duplicate_mode" type="C" size="2" />
     <field name="redirect_on_duplicate" type="C" size="1024" />
+    <field name="sp_type" type="C" size="1" />
```
boshrin, line 42: For consistency with other enums, this should be size 2.

app/Plugin/EnvSource/Lib/enum.php
```diff
@@ -31,3 +31,8 @@ class EnvSourceDuplicateModeEnum
   const LoginIdentifier = "LI";
   const SORIdentifier   = "SI";
 }
+
+class AuthProviderEnum
+{
+  const Shibboleth   = ";";
```
boshrin, line 37: This is going to make SQL query construction really annoying :) The enum const should be something like "SH" (for "Shibboleth").

We should also define "Other" (code "O" for consistency") so if any deployer is using something else they don't get confused. We could proactively define "SimpleSamlPHP" ("SS"?) as well even though we're not doing anything specific for it.

app/Plugin/EnvSource/Lib/lang.php
```diff
@@ -42,7 +42,11 @@
     EnvSourceDuplicateModeEnum::AnyIdentifier    => 'Any Identifier Match',
     EnvSourceDuplicateModeEnum::LoginIdentifier  => 'Login Identifier Match',
   ),
-  
+
+  'pl.envsource.en.auth.provider' => array(
+    AuthProviderEnum::Shibboleth    => 'Shibboleth',
```
boshrin, line 47: For pedantic reasons, this should probably say "Shibboleth SP".

app/Plugin/EnvSource/Lib/lang.php
```diff
@@ -52,8 +56,9 @@
   'er.envsource.token'          => 'Token error',
   
   // Plugin texts
-  'pl.envsource.mode.dupe'      => 'Duplicate Handling Mode',
-  'pl.envsource.name.unknown'   => 'Unknownname',
-  'pl.envsource.redirect.dupe'  => 'Redirect on Duplicate',
-  'pl.envsource.sorid.desc'     => 'This must be set to an environment variable holding a unique identifier for any authenticated user.'
+  'pl.envsource.mode.dupe'               => 'Duplicate Handling Mode',
+  'pl.envsource.sp.type'                 => 'Multi Value Delimiter',
```
boshrin, line 60: The goal here is to document which SP is used for the integration, not the behavior of that SP. So this field should be labeled something like "Web Server SP Provider".

boshrin (changes requested):

app/Plugin/EnvSource/Config/Schema/schema.xml
ioigoume: Done

app/Plugin/EnvSource/Lib/lang.php
ioigoume: Done

app/Plugin/EnvSource/Lib/lang.php
ioigoume: Done

app/Plugin/EnvSource/Lib/enum.php
ioigoume: Done

app/Plugin/EnvSource/Config/Schema/schema.xml
```diff
@@ -39,6 +39,8 @@
     </field>
     <field name="duplicate_mode" type="C" size="2" />
     <field name="redirect_on_duplicate" type="C" size="1024" />
+    <field name="sp_type" type="C" size="2" />
+    <field name="other_type" type="C" size="2" />
```
boshrin, line 43: I don't think we need `other_type`, see additional comment below.

app/Plugin/EnvSource/Model/EnvSourceBackend.php
```diff
@@ -126,9 +126,34 @@ protected function resultToOrgIdentity($result) {
...
-      $orgdata['EmailAddress'][0]['type'] = EmailAddressEnum::Official;
-      $orgdata['EmailAddress'][0]['verified'] = true;
+      if(!empty($this->pluginCfg['sp_type'])) {
+        $delimiter = "";
+        if($this->pluginCfg['sp_type'] == AuthProviderEnum::Shibboleth) {
+          $delimiter = ";";
+        } elseif($this->pluginCfg['sp_type'] == AuthProviderEnum::Simplesamlphp) {
+          $delimiter =",";
+        } elseif($this->pluginCfg['sp_type'] == AuthProviderEnum::Other) {
+          $delimiter =  trim($this->pluginCfg['other_type']);
```
boshrin, line 136: If type is `Other` we simply don't support multiple values, and we don't try to `explode()` the value. This also simplifies the user interface, below.

boshrin (changes requested):

app/Plugin/EnvSource/Model/EnvSourceBackend.php
ioigoume: Done

app/Plugin/EnvSource/Config/Schema/schema.xml
ioigoume: Done

PR #542, opened 2023-09-21
CO-2699_UI_prevents_COU_admin_for_child_COU_from_deleting_role_in_the_child_COU

ioigoume:


boshrin (changes requested):
Aside from the generic name of the Helper, there are some anti-patterns here. Specifically:

1. Views shouldn't call Models.
2. This Helper call appears to be to assist in a permissions calculation. This should be done in the Controller instead.

Specifically, `$permissions['cous']` is supposed to contain the list of COUs that the current Admin can manage. If we need to run the tree logic to expand that list, `CoPeopleController::isAuthorized()` should be updated to correctly populate that variable instead.

PR #550, opened 2023-10-30
CO-2712_CoEnrollmentAttributeDefault_validateModified_can_not_handle_save_from_CMD

ioigoume:


boshrin:
We should just put the JIRA number in the comment so when we look at that later we know why it's there.

PR #552, opened 2023-11-09
CO-2715_Enable_safeties_save_option_for_Oauth2Server_Model

ioigoume:


boshrin:
Clever reuse of the `safeties` parameter that was never actually used anywhere. OK to merge.

PR #555, opened 2023-11-23
Add port config for SqlServer (CO-2710)

skoranda:


boshrin:
I'm generally OK with this, but have we tested against Postgres?

skoranda:
Yes. It works.

PR #563, opened 2023-12-08
CO-2732_Improve_LDAP_Provisioner_plugin_logging

ioigoume:


ioigoume:
> Why are we defining an enum with all the LDAP error codes defined? This seems like something we shouldn't have to maintain...
> 
> And what is `en.ldapprovisioner.commonerrors` for? Why can't we use `ldap_err2str()` (which is what `logLdapError()` seems to do anyway)?

@boshrin i removed the lang.php entry and for the enum class i only kept the codes that are currently used by the plugin as part of if conditions. If we need more codes to check against in the future, we can extend the list. To my opinion, by having the codes as part of an enumerator class improves the code readability. Also we only keep the codes at one place instead of copy and pasting the actual ldap code number.

PR #564, opened 2023-12-10
CO-2733_SqlProvisioner_assumes_path_always_under_LOCAL

ioigoume:


boshrin:
See comment in CO-2733.

ioigoume:
> See comment in CO-2733.

@boshrin added comment in CO-2733

PR #539, opened 2023-09-02
CO-2697_Add_missing_model_associations._Changelog_behavior_to_support_contain_equal_to_true

ioigoume:
- Add missing associations Pipeline,CoEnrollmentFlow,CoGroup,DataFilter
- Changelog Behavior enable correct find for contain equals to true.

PR #576, opened 2024-02-05
CO-2675_UnixCluster_plugin_does_not_use_Primary_Name_for_GECOS

ioigoume:


boshrin:
I'm confused by the need to change

```
'CoPersonRole' => array('CoPerson' => 'PrimaryName')
```

to

```
'CoPersonRole' => array('CoPerson' => array(
  'PrimaryName' => array('conditions' => array('PrimaryName.primary_name' => true))
))
```

`PrimaryName` is defined as a relation on both `CoPerson` and `OrgIdentity` so that related queries automatically pick up the conditions necessary to filter on primary_name=true without having to respecify them in each related model. Is something breaking that?

ioigoume:
> I'm confused by the need to change
> 
> ```
> 'CoPersonRole' => array('CoPerson' => 'PrimaryName')
> ```
> 
> to
> 
> ```
> 'CoPersonRole' => array('CoPerson' => array(
>   'PrimaryName' => array('conditions' => array('PrimaryName.primary_name' => true))
> ))
> ```
> 
> `PrimaryName` is defined as a relation on both `CoPerson` and `OrgIdentity` so that related queries automatically pick up the conditions necessary to filter on primary_name=true without having to respecify them in each related model. Is something breaking that?

This Pull Request is related to [PR-539](https://github.com/Internet2/comanage-registry/pull/539) that has been already merged. I will retest/clean up the fix and follow up on the question/comment

boshrin:
Ioannis' comment from Slack:

> i revisited [CO-2675](https://github.com/Internet2/comanage-registry/pull/576) pull request and the reason why the PrimaryName virtual class does not always work is the contain . If we build a query but do not add our own models in the contain then it works. If we do, it will not. This is why the PR mainly changes the edit/view/delete_contains arrays. It also changes some individual queries that try to filter the records using the PrimaryName. So far, i could not link the issue with the ChangelogBehavior.

So this issue probably isn’t ChangelogBehavior, but something with Containable, which puts us into Cake v2 core, which we probably don’t want to touch if we can avoid it.
This probably falls into the category of “unsatisfying, but not worth the effort to fix”.

PR #595, opened 2024-05-10
CO-2178_Plugin_connection_with_ORCID_member_API

ioigoume:


lpaglione:
@boshrin I reviewed the functionality side of this pull request, and it looks good.  I think two minor improvements should be considered. These were not specified in the original scope written in CO-2178, though they should be considered before the v4.0.0 release. They are:

**Support the Storage of Multiple ORCID Access Tokens**
The ORCID API allows for permissions (access tokens) to be given to a client for an ORCID iD multiple times with different scopes - for example, one access token will provide just the read-limited scope; another may provide a couple of update scopes. These are important because when obtaining permissions from the user for different reasons (or over time), you may need to change the scopes requested. In addition, the ORCID OAuth flow will not ask the user again for permissions that have already been granted UNLESS that permission had been revoked by the user. In both of these use cases (different scopes / access revocation) an additional access token would need to be stored in the new orcid token table for the iD/API Client pair. At the moment, the plugin will only allow for one entry in the table.

**Inclusion of the "create" date stamp in the orcid_token API response**
This information will be particularly important to complement the information stored in the orcid token table when multiple access tokens can be stored - this will be how folks using the ORCID API client can determine the freshest access token when there is more than one with the same scope (for example, if the user had revoked and reissued permissions).

Again, these two items do not prevent the overall plugin from doing the explicit items outlined in CP-2178

PR #611, opened 2024-06-13
Fix "Possible Matches" advisory box to display on three characters (CO-2777)

spartanicus:
This PR also improves the box's layout and style.

image

### DIFF
--- a/app/Controller/CoPeopleController.php
+++ b/app/Controller/CoPeopleController.php
@@ -1109,7 +1109,7 @@ class CoPeopleController extends StandardController {
       return;
     }
     // Validate the request data
-     list($criteria, $invalidFields, $unProcessedFields) = $this->CoPerson->validateRequestData($reqData);
+    list($criteria, $invalidFields, $unProcessedFields) = $this->CoPerson->validateRequestData($reqData);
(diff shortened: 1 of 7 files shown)


ioigoume:
Hi @spartanicus ,
the PR looks great. I have one fix and one suggestion:

## Fix

In the `CoPerson::match` method
```php
    if(!empty($args)) {
      $args['conditions']['PrimaryName.primary_name'] = true;
      $args['conditions']['CoPerson.co_id'] = $coId;
      if($mdl == 'Name') {
        $args['conditions']['Name.primary_name'] = true;
      }
      $args['contain'][] = 'PrimaryName';
      $args['contain'][] = 'CoPersonRole';
    } else {
      return [];
    }
```

## Suggestion

In the `View/CoPeople/match.ctp` print the Person Id as we do for all other use cases, e.g. for the PeoplePicker.

```php
        print $this->Html->link(
            generateCn($m['PrimaryName']) . ' (' . $m['CoPerson']['id'] . ')',
            array('controller' => 'co_people', 'action' => 'canvas', $m['CoPerson']['id'])
        );
```

ioigoume (approved):
LGTM

PR #624, opened 2024-07-05
CO-2797_Changing_Minimum_value_for_Identifier_Assignment_does_not_work

ioigoume:


shaynasings:
Question - if the configuration is changed for the sequential identifier assignment minimum to a lower number than it was previously,  should the new values still start with whatever the previous (higher) minimum was? 

In other words, this pull request only works if the admin changes the sequential identifier assignment minimum to a higher value. If the new minimum is lower than the previous one, it seems like the assignments should start with the new, lower number.  

This can be addressed as Mike mentioned by deleting the identifier assignment configuration and recreating it with the lower minimum. This will create new identifier assignments starting at the new lower minimum,  even if there are remaining identifiers on CO Person records of the associated type with higher values.   Luckily it appears that Registry will skip those numbers as they come up. However, I think Mike's point was that you shouldn't have to delete the identifier assignment to change the minimum, not matter whether it's higher or lower.

ioigoume:
> Question - if the configuration is changed for the sequential identifier assignment minimum to a lower number than it was previously, should the new values still start with whatever the previous (higher) minimum was?
> 
> In other words, this pull request only works if the admin changes the sequential identifier assignment minimum to a higher value. If the new minimum is lower than the previous one, it seems like the assignments should start with the new, lower number.
> 
> This can be addressed as Mike mentioned by deleting the identifier assignment configuration and recreating it with the lower minimum. This will create new identifier assignments starting at the new lower minimum, even if there are remaining identifiers on CO Person records of the associated type with higher values. Luckily it appears that Registry will skip those numbers as they come up. However, I think Mike's point was that you shouldn't have to delete the identifier assignment to change the minimum, not matter whether it's higher or lower.

@shaynasings That was my initial thinking and the initial implementation had a database change that facilitated exactly this behavior. However, following up on Benn's comment, in Slack, I updated the change to this one. Now it works like a sequential column of a database. Which means it can only go up and even if you update the configuration to a smaller value then it will still go up.

shaynasings:
Ok - then @boshrin this works as Ioannis describes above in my testing

PR #637, opened 2024-08-27
Provision CoGroup after UnixCluster assign (CO-2820)

skoranda:


boshrin:
There's a lot of whitespace noise in the diff that makes it hard to see what the actual change is. It would be better to split that out so it's easier to read. That said, the actual change seems ok, assuming you've tested it out and it works for you.

Note this PR is against hotfix-4.3.x, and the next release is expected to be 4.4.0 from develop. I'll leave it to @shaynasings to address that.

skoranda:
That's why there are two commits. The first one explicitly reads "Fix numerous tab issues". You can see the diff just for the second commit.

boshrin:
Sorry, I jumped straight to "Files Changed" and missed that the commits tab was showing two.

PR #638, opened 2024-08-28
CO-2816_Present_Subset_of_COUs_in_Enrollment_Flow

ioigoume:


boshrin:
A couple of minor comments…

I think we should call this “Configuration Labels” instead of just “Labels”. It’s a more descriptive term and also constrains its scope a bit.

If color is going to be defined as a hex value (which seems fine to me) then the field should be defined as `varchar(6)` (or maybe `varchar(8)`) to make it harder to store the wrong type of value. Also, we should just decide whether the leading `#` is required or omitted so we don’t end up with both formats.

ioigoume:
> A couple of minor comments…
> 
> I think we should call this “Configuration Labels” instead of just “Labels”. It’s a more descriptive term and also constrains its scope a bit.
> 
> If color is going to be defined as a hex value (which seems fine to me) then the field should be defined as `varchar(6)` (or maybe `varchar(8)`) to make it harder to store the wrong type of value. Also, we should just decide whether the leading `#` is required or omitted so we don’t end up with both formats.

Addressed both of them. Model renamed and the validation rule has been modified in order to ensure the hex string consistency.

app/Config/Schema/schema.xml
```diff
@@ -44,7 +44,37 @@
...
+  <table name="configuration_labels">
+    <field name="id" type="I">
+      <key />
+      <autoincrement />
+    </field>
+    <field name="co_id" type="I">
+      <constraint>REFERENCES cm_cos(id)</constraint>
+    </field>
+    <field name="label" type="C" size="64" />
+    <field name="color" type="C" size="8" />
```
spartanicus, line 57: The color picker can include an alpha channel resulting in strings such as #a33937a4 - so size="8" is not enough.

ioigoume: Thank you @spartanicus . I fixed the bug

app/Model/ConfigurationLabel.php
```diff
@@ -0,0 +1,164 @@
...
+   *   The hexadecimal string to validate. Must contain a leading '#'
+   *
+   * @return bool TRUE if $hex is valid or FALSE if it is not.
+   * @since  COmanage Registry v4.4.0
+ */
+  public function validateHex($check) {
+    if (!is_string($check['color'])) {
+      return false;
+    }
+    return preg_match('/^(#[0-9a-fA-F]{6})$/', $check['color']) === 1;
```
spartanicus, line 147: {6} needs to become {8} to allow for the optional alpha channel

PR #641, opened 2024-08-30
Better scaling for petition index view (CO-2819)

skoranda:


skoranda:
I am going to close this pull request and submit another that uses a more flexible solution and is a merge targeted for develop instead of hotfix-4.3.x.

PR #645, opened 2024-09-16
CO-2826_Provisioned_Services_page_times_out

ioigoume:


boshrin:
Is there a reason we need a configuration on the LDAP Provisioner that adds complexity and is unlikely to get used? Shouldn't we just set a reasonable default?

Also, if the problem is that the provisioned services page doesn't render, this won't prevent the problem from happening if a different type of plugin takes too long to respond. We should either have the code that calls the plugin timeout, or review all provisioning plugins and make sure they timeout correctly (and add a note to the developer's manual.)

(I'm assuming the use case here is that the LDAP server is down for maintenance or something. If it's being decommissioned then the correct solution is of course for the admin to remove it from the configuration.)

ioigoume:
@boshrin ,

> Is there a reason we need a configuration on the LDAP Provisioner that adds complexity and is unlikely to get used? Shouldn't we just set a reasonable default?

I agree. I just took the long way since we have a minor release ahead of us which allow database changes. Then it will be the administrators responsibility to handle it appropriately

> Also, if the problem is that the provisioned services page doesn't render, this won't prevent the problem from happening if a different type of plugin takes too long to respond. We should either have the code that calls the plugin timeout, or review all provisioning plugins and make sure they timeout correctly (and add a note to the developer's manual.)

This is true. There has not been a report for the other plugins so far. I could create a catch all ticket it we need to.

> (I'm assuming the use case here is that the LDAP server is down for maintenance or something. If it's being decommissioned then the correct solution is of course for the admin to remove it from the configuration.)

Correct. The timeout will allow the view to render and not timeout, as it does at the moment. The view will report the issue back to the administrator and it will be easier to fix. At the moment, there is no feedback, which makes administrator's work more challenging.

PR #667, opened 2024-12-05
Provide "Do not close your browser" message on enrollment flow redirect page (CO-2844)

spartanicus:
This PR adds a simple message above the redirect information box during an enrollment flow that looks like this:
image


ioigoume (approved):
LGTM

PR #668, opened 2024-12-05
Make top global search significantly more obvious. (CO-2846)

spartanicus:
This PR pulls the top search out of a dropdown and places it to the left near the hamburger menu. The work required reworking the topbar using more modern practices (such as flexbox). The result looks like this:
image

### DIFF
--- a/app/View/CoDashboards/search.ctp
+++ b/app/View/CoDashboards/search.ctp
@@ -58,7 +58,13 @@
   <div id="search">
     <?php
       print $this->Form->input('q', $options);
-      print $this->Form->submit(_txt('op.search'));
+      $options = array(
+        'type' => 'submit',
+        'class' => 'btn btn-primary global-search-submit',
+        'escape' => false
+      );
+      $submitButtonText = '<span class="material-icons-outlined">search</span><span class="search-button-text">' . _txt('op.search') . '</span>';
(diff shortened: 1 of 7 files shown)


ioigoume:
@spartanicus The change overall looks good. I found a small thing that might need fixing. It is not part of this fix that it is very related and i came across it while testing. See the screenshot:
Screenshot from 2025-01-18 16-44-13

ioigoume (changes requested):
Please check the comment

spartanicus:
I had intentionally left that button large on mobile because it makes it easier to click - however, if you think this looks unintentional (i.e. broken), I'll give it a change. Here's the current look on iPhone 11:
image

ioigoume:
> I had intentionally left that button large on mobile because it makes it easier to click - however, if you think this looks unintentional (i.e. broken), I'll give it a change. Here's the current look on iPhone 11: image

Since this is intentional then it works for me

ioigoume (approved):
LGTM

spartanicus:
I've updated it to look less peculiar - it was a good opportunity to clean up the search page's form a bit more. I agree that it didn't look good, even if intentional. (And when someone looks at an interface and the first thought that comes to mind is 'broken' that's an important flag to pay attention to, even if it doesn't always mean we should change it. ...But it often does mean that. ;-) 
image

PR #669, opened 2024-12-06
Add Enrollment Flow Description field. (CO-2849)

spartanicus:
This addition provides an Enrollment Flow description on the Enrollment Flow index and on the "Select Enrollment Flow" listing.

image

image

image

### DIFF
--- a/app/Model/CoEnrollmentFlow.php
+++ b/app/Model/CoEnrollmentFlow.php
@@ -120,9 +120,14 @@ class CoEnrollmentFlow extends AppModel {
       'message' => 'A name must be provided'
     ),
     'description' => array(
-      'rule' => 'notBlank',
-      'required' => false,
-      'allowEmpty' => true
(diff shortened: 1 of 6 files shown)


app/Model/CoEnrollmentFlow.php
```diff
@@ -119,6 +119,11 @@ class CoEnrollmentFlow extends AppModel {
       'required' => true,
       'message' => 'A name must be provided'
     ),
+    'description' => array(
```
ioigoume, line 122: this needs to be:
```php
    'description' => array(
      'content' => array(
        'rule' => array('maxLength', 256),
        'required' => false,
        'allowEmpty' => true
      ),
      'filter' => array(
        'rule' => array('validateInput')
      )
    ),
```

spartanicus: Fixed!

PR #671, opened 2024-12-20
UnixCluster posixAccount posixGroup LdapProvisioner expunge (CO-2869)

skoranda:


boshrin (commented):
Per comments in CO-2602, this would be clearer as an initialization rather than a dereferencing. eg something like

```
$attributes = array();
$attributes['objectclass'] = array();
```

all the way back up around line 268.

PR #681, opened 2025-01-16
Enable delete all locks (CO-2890)

skoranda:


skoranda:
I disagree. The combination of "deleteAll" and 'TRUE' could not be more self-documenting, whereas passing in an array that makes assumptions (correct in this case) about whether or not a column can be null requires knowledge about the code base outside of the statement.

Nevertheless, to not waste anymore time on a trivial code update that provides useful functionality for those that operate Registry in production I have made the change.

PR #692, opened 2025-02-17
CO-2774_Internal_Error_when_linking_an_Org_Identity_to_a_CO_Person

ioigoume:


boshrin:
I'm not comfortable changing the default HTTP timeout to 5 seconds. Match requests can easily take greater than 5 seconds (with large datasets and complex match rules), meaning our out of the box configuration would break for certain categories of deployments.

There are also a number of other unrelated uses of CoHttpClient that may have longer transaction times, especially (eg) callouts to remote API services.

Is reducing the timeout really required for this patch?

ioigoume:
@boshrin ,
thank you for referring to the use cases. I reverted the timeout change.

PR #701, opened 2025-03-07
Add "About this Group" introduction field to co_group (CO-2847)

spartanicus:


spartanicus:
This is a simple PR that adds an introduction field to groups labelled "About this group"
image

PR #713, opened 2025-03-25
CO-2910_Duplicate_Enrollment_flow_does_not_copy_the_Enrollment_Flow_Wedge_Configuration

ioigoume:


app/AvailablePlugin/MeemEnroller/Model/MeemEnroller.php
```diff
@@ -67,7 +67,11 @@ class MeemEnroller extends AppModel {
       'foreignKey' => 'mfa_exempt_co_group_id'
     )
   );
-  
+
+  public $hasMany = array(
+    'MeeMfaStatus' => array('dependent' => true),
```
boshrin, line 72: "Mee<ins>m</ins>MfaStatus" ?

app/Model/AppModel.php
```diff
@@ -849,7 +849,184 @@ public function delete($id = null, $cascade = true) {
...
+   * @param  String  $foreignKey The name of the foreign key from $model to its parent (or 'id' if no parent)
+   * @param  Integer $fkid       The ID of the foreign key in $model
+   * @param  Array   $idmap      Reference to an array of old IDs to new IDs (originals to duplicates)
+   * @param  Boolean $isTree     True if $model implements TreeBehavior
+   * @param  Boolean $alterName  if true we will change the name and description of the copied model.
+   * @return Boolean True on success
+   * @throws InvalidArgumentException
+   * @throws RuntimeException
+   */
+
```
boshrin, line 867: I'm not a huge fan of adding even more stuff to AppModel, but I suppose in v4 that's our least bad option. (In PE we'd probably want to use a trait instead.)

app/Model/AppModel.php
```diff
@@ -849,7 +849,184 @@ public function delete($id = null, $cascade = true) {
...
+  /**
+   * Duplicate objects.
+   *
+   * @since  COmanage Registry v3.2.0
+   * @param  Model   $model      Model to duplicate
+   * @param  String  $foreignKey The name of the foreign key from $model to its parent (or 'id' if no parent)
+   * @param  Integer $fkid       The ID of the foreign key in $model
+   * @param  Array   $idmap      Reference to an array of old IDs to new IDs (originals to duplicates)
+   * @param  Boolean $isTree     True if $model implements TreeBehavior
```
boshrin, line 861: I realize this is just moving code old around, but we probably should have done something like `$model->Behaviors->loaded('Tree')` to test for TreeBehavior instead of having to explicitly pass a parameter...

app/Model/CoEnrollmentFlow.php
```diff
@@ -847,85 +847,87 @@ public function configuredSteps($id) {
   /**
    * Duplicate an existing Enrollment Flow.
    *
-   * @since  COmanage Registry v0.9.2
```
boshrin, line 850: The implementation may be changing but the function was still introduced in v0.9.2.

app/Model/CoEnrollmentFlow.php
```diff
@@ -847,85 +847,87 @@ public function configuredSteps($id) {
...
-    $ef = $this->find('first', $args);
-    
-    if(empty($ef)) {
+    // This maintains a per-Model map of old_id -> new_id
+    $idmap = array();
+
+    // (0) Start a transaction. If we error out, we can roll back.
+    $this->_begin();
+
+    // (1) First copy the CO itself
```
boshrin, line 864: We're copying the Enrollment Flow here, not the CO?

boshrin (changes requested):

app/AvailablePlugin/MeemEnroller/Model/MeemEnroller.php
ioigoume: Done. Good catch

app/Model/AppModel.php
ioigoume: With this change we keep the duplicate Object core functionality in one place.

app/Model/AppModel.php
ioigoume: Done

app/Model/CoEnrollmentFlow.php
ioigoume: Done

app/Model/CoEnrollmentFlow.php
ioigoume: done

ioigoume:
@boshrin review comments addressed.

PR #729, opened 2025-05-06
CO-2928 Add AWS DynamoDB session handler plugin

rmblair:


app/Config/core.php
```diff
@@ -172,9 +172,36 @@
...
  */
-	Configure::write('Session', array(
-		'defaults' => 'php'
-	));
+
+    // AWS DynamoDB sessions
+    if (getenv('COMANAGE_REGISTRY_DYNAMODB_REGION') and
+        getenv('COMANAGE_REGISTRY_DYNAMODB_PHPSESSIONS_TABLE') and
+        getenv('COMANAGE_REGISTRY_DYNAMODB_PHPSESSIONS_ACCESSKEY') and
+        getenv('COMANAGE_REGISTRY_DYNAMODB_PHPSESSIONS_SECRETACCESSKEY')) {
```
boshrin, line 180: Is there a reason to use `and` instead of `&&` here?

See eg [this discussion](https://stackoverflow.com/questions/2803321/and-vs-as-operator).

app/Config/core.php
```diff
@@ -172,9 +172,36 @@
...
+                // session.gc_maxlifetime in seconds
+                // default is 1440 (24 minutes)
+                'session.gc_maxlifetime' => 86400,
+            )
+        ));
+    } else {
+        // default session config
+        Configure::write('Session', array(
+            'defaults' => 'php',
+            'cookie' => 'CAKEPHP',
```
boshrin, line 202: This might be a bit of an academic question since Cake 2 isn't getting any further updates, but is there a reason to explicitly set the `cookie` name here rather than let Cake set a default name?

rmblair: I believe I reused what CakePHP was shipping as a default cookie name in app/Config/core.php, though post-v2.x [this discussion](https://github.com/cakephp/cakephp/issues/11584) occurred in their issue tracker and they decided to drop the default in v3.6. Let me know if you would prefer that I add another commit with that change.

PR #740, opened 2025-07-17
CO-2501_Manual_Provisioning_Can_Sometimes_Happen_Automatically

ioigoume:


app/Model/Behavior/ProvisionerBehavior.php
```diff
@@ -29,6 +29,11 @@
 App::uses('CakeSession', 'Model/Datasource');
 
 class ProvisionerBehavior extends ModelBehavior {
+  
+  // Auxiliary property that faciliates manual provisioning
+  // Set once per request
+  private $isRequestToProvision = false;
```
boshrin, line 35: Since this is just a flag passed between requestToProvision and manualProvision, why not just pass it as a flag?

Also (and I admit this is somewhat arbitrary) it seems like requestToProvision should be the "core" function and manualProvision should be the wrapper (in which case the flag can simply be a `$manual=false` parameter added to requestToprovision).

ioigoume: I renamed the `manualProvision` to` requestToProvision` and made the `$manual` a flag

PR #742, opened 2025-08-04
CO-2939_Storing_the_entityId_through_an_Enrollment_Flow_shows_only_the_organization_id

ioigoume:


jensalmer:
This fixes the issue as I found it. The enrollment form now displays the organization name with the id in parenthesis instead of it's id. The petition review also now shows the same format.

PR #746, opened 2025-09-18
NO_TICKET_fix_error_message

ioigoume:


boshrin (commented):
There are two places where `er.ois.linked` is used. If this is a better error message, why aren't we just changing the default localization (which a deployer could also just change themselves) instead of splitting apart the errors into two separate language keys?

rshiggin:
> There are two places where `er.ois.linked` is used. If this is a better error message, why aren't we just changing the default localization (which a deployer could also just change themselves) instead of splitting apart the errors into two separate language keys?

The error key is being used for different errors. We need to associate at least one of these with a different key to configure separate error messages in our localization.

rshiggin:
Tested successfully w/ 4.5.1 (Indiana University). Thanks!

boshrin (approved):

PR #757, opened 2026-04-20
CO-2977_Name_Make_Primary_intermittently_fails_with_Unknown_CO_Person

ioigoume:


boshrin:
I assume we can disable provisioning here because it fires later?

ioigoume:
> I assume we can disable provisioning here because it fires later?

@boshrin 
I updated the comment to better reflect the purpose of the change, which comes down to two points: Provisioning is skipped here because this saveField is an intermediate step inside beforeSave() — at this point the new primary name hasn't been saved yet, so the record would temporarily have no PrimaryName, causing provisioning lookups to fail It's safe to skip because the outer save() will still trigger afterSave() on the ProvisionerBehavior once the full operation completes, ensuring everything gets provisioned correctly. Also confirmed correct provisioning behavior through manual testing.

PR #759, opened 2026-05-14
CO-2985_CoPerson_Indexed_Slowed_By_Large_Number_of_COUs

ioigoume:
Cache cou queries to improve performance
- total number of queries dropped from approximately 240 to approximately 140 by request for the MyPopulation views.


boshrin:
The original reporter has a workaround, so this effort is on hold.

